PR #5, opened 2023-10-06
Update MerlinAutoUpdate.sh

Martinski4GitHub:
Fixed code to get router model ID. Other code improvements.

### DIFF
--- a/MerlinAutoUpdate.sh
+++ b/MerlinAutoUpdate.sh
@@ -3,13 +3,13 @@
 URL_BASE="https://sourceforge.net/projects/asuswrt-merlin/files"
 URL_RELEASE_SUFFIX="Release"
 
-##-------------------------------------##
-## Added by Martinski W. [2023-Oct-05] ##
-##-------------------------------------##
+##----------------------------------------------##
+## Added/Modified by Martinski W. [2023-Oct-05] ##
+##----------------------------------------------##
 _GetRouterModel_()
 {
    local retCode=1  routerModelID=""
-   local nvramModelKeys="odmpid productid wps_modelnum wps_device_name model"
+   local nvramModelKeys="productid build_name odmpid"
    for nvramKey in $nvramModelKeys
    do
        routerModelID="$(nvram get "$nvramKey")"
@@ -26,6 +26,12 @@ SETTINGSFILE="$SETTINGS_DIR/custom_settings.txt"
 # Define the cron schedule and command to execute
 CRON_SCHEDULE="0 0 * * 0"
 COMMAND="sh /jffs/scripts/MerlinAutoUpdate.sh run_now"
+
+##-------------------------------------##
+## Added by Martinski W. [2023-Oct-05] ##
+##-------------------------------------##
+_WaitForEnterKey_()
+{ printf "\nPress Enter to continue..." ; read EnterKEY ; }
 
 # Function to get LAN IP
 get_lan_ip() {
@@ -129,23 +135,28 @@ Update_Custom_Settings() {
 }
 
 get_current_firmware() {
-    local current_version="$(nvram get buildno)_$(nvram get extendno)"
+    local current_version="$(nvram get buildno).$(nvram get extendno)"
     echo "$current_version"
 }
 
+##----------------------------------------##
+## Modified by Martinski W. [2023-Oct-05] ##
+##----------------------------------------##
 get_latest_firmware() {
     local url="$1"
-    local page=$(curl -s "$url")
 
-    local links_and_versions=$(echo "$page" | grep -o 'href="[^"]*'"$MODEL"'[^"]*\.zip' | sed 's/amp;//g; s/href="//' | 
-        awk -F'[_\.]' '{print $3"."$4"."$5" "$0}' | sort -t. -k1,1n -k2,2n -k3,3n)
+    local links_and_versions="$(curl -s "$url" | grep -o 'href="[^"]*'"$MODEL"'[^"]*\.zip' | sed 's/amp;//g; s/href="//' | 
+        awk -F'[_\.]' '{print $3"."$4"."$5" "$0}' | sort -t. -k1,1n -k2,2n -k3,3n)"
 
-    local latest=$(echo "$links_and_versions" | tail -n 1)
-    local version=$(echo "$latest" | cut -d' ' -f1 | awk -F. '{print $(NF-1)"."$NF}')
-    local link=$(echo "$latest" | cut -d' ' -f2-)
+    if [ -z "$links_and_versions" ]
+    then echo "**ERROR** **NO_URL**" ; return 1 ; fi
+
+    local latest="$(echo "$links_and_versions" | tail -n 1)"
+    local version="$(echo "$latest" | cut -d' ' -f1 | awk -F. '{print $(NF-1)"."$NF}')"
+    local link="$(echo "$latest" | cut -d' ' -f2-)"
 
     # Extracting the correct link from the page
-    local correct_link=$(echo "$link" | sed 's|^/|https://sourceforge.net/|')
+    local correct_link="$(echo "$link" | sed 's|^/|https://sourceforge.net/|')"
 
     echo "$version"
     echo "$correct_link"
@@ -196,7 +207,7 @@ pure_file=$(ls | grep -i '_pureubi.w' | grep -iv 'rog')
         Update_Custom_Settings "local" "n"
     fi
 
-    read -p "Press Enter to continue..."
+    _WaitForEnterKey_
 }
 
 # Function to translate cron schedule to English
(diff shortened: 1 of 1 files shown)


MerlinAutoUpdate.sh
```diff
@@ -3,13 +3,13 @@
...
-## Added by Martinski W. [2023-Oct-05] ##
-##-------------------------------------##
+##----------------------------------------------##
+## Added/Modified by Martinski W. [2023-Oct-05] ##
+##----------------------------------------------##
 _GetRouterModel_()
 {
    local retCode=1  routerModelID=""
-   local nvramModelKeys="odmpid productid wps_modelnum wps_device_name model"
+   local nvramModelKeys="productid build_name odmpid"
```
Firetop, line 12: Thank you for this!

Good change to priority order of the NVRAM keys used and removing the extra keys.

MerlinAutoUpdate.sh
```diff
@@ -27,6 +27,12 @@ SETTINGSFILE="$SETTINGS_DIR/custom_settings.txt"
 CRON_SCHEDULE="0 0 * * 0"
 COMMAND="sh /jffs/scripts/MerlinAutoUpdate.sh run_now"
 
+##-------------------------------------##
+## Added by Martinski W. [2023-Oct-05] ##
+##-------------------------------------##
+_WaitForEnterKey_()
+{ printf "\nPress Enter to continue..." ; read EnterKEY ; }
+
```
Firetop, line 35: Cleaner than all the prints on the script, this is approved as well. :)

MerlinAutoUpdate.sh
```diff
@@ -128,26 +134,40 @@ Update_Custom_Settings() {
     esac
 }
 
+##----------------------------------------##
+## Modified by Martinski W. [2023-Oct-05] ##
+##----------------------------------------##
 get_current_firmware() {
-    local current_version="$(nvram get buildno)_$(nvram get extendno)"
+    local current_version="$(nvram get buildno).$(nvram get extendno)"
```
Firetop, line 141: Will need to review implications more with further audit of dependencies down the script:

Version Format: If other parts of the script or system expect the version to be in the format of buildno_extendno (e.g., 1234_5678), the modified version will break that expectation. Conversely, if the system or a new standard expects the version to be in the format of buildno.extendno (e.g., 1234.5678), the modification is correct.

MerlinAutoUpdate.sh
```diff
@@ -445,7 +500,7 @@ else
     Say "Login failed. Please confirm credentials by selecting: 1. Configure Credentials"
 fi
 
-read -p "Press Enter to continue..."
+    _WaitForEnterKey_
```
Firetop, line 503: Approved.

MerlinAutoUpdate.sh
```diff
@@ -507,7 +562,7 @@ while true; do
     4) change_schedule ;;
     e) exit ;;
     *) echo "Invalid choice. Please try again."
-       read -p "Press Enter to continue..."  # Pauses script until Enter is pressed
+       _WaitForEnterKey_  # Pauses script until Enter is pressed
```
Firetop, line 565: Approved.

MerlinAutoUpdate.sh
```diff
@@ -196,7 +216,7 @@ pure_file=$(ls | grep -i '_pureubi.w' | grep -iv 'rog')
         Update_Custom_Settings "local" "n"
     fi
 
-    read -p "Press Enter to continue..."
+    _WaitForEnterKey_
```
Firetop, line 219: Approved.

MerlinAutoUpdate.sh
```diff
@@ -284,22 +304,57 @@ else
...
 set -- $(get_latest_firmware "$URL_RELEASE")
-release_version=$1
-release_link=$2
+release_version="$1"
+release_link="$2"
+
+    if [ "$1" = "**ERROR**" ] && [ "$2" = "**NO_URL**" ] 
+    then
+        echo "**ERROR**: No firmware release URL was found for [$MODEL] router model."
+        _WaitForEnterKey_
```
Firetop, line 348: Approved.

MerlinAutoUpdate.sh
```diff
@@ -128,26 +134,40 @@ Update_Custom_Settings() {
...
+    local current_version="$(nvram get buildno).$(nvram get extendno)"
     echo "$current_version"
 }
 
+##----------------------------------------##
+## Modified by Martinski W. [2023-Oct-05] ##
+##----------------------------------------##
 get_latest_firmware() {
     local url="$1"
-    local page=$(curl -s "$url")
```
Firetop, line 138: Direct processing without storing in a variable can be more efficient, especially if the data is used only once. It reduces memory usage (although for a small script like this the impact is negligible). However, if debugging is required, having the data in a separate variable can be helpful. For the purpose of this script, the modification seems logical for now, debugging this step should no longer be required.

MerlinAutoUpdate.sh
```diff
@@ -128,26 +134,40 @@ Update_Custom_Settings() {
...
 
-    local links_and_versions=$(echo "$page" | grep -o 'href="[^"]*'"$MODEL"'[^"]*\.zip' | sed 's/amp;//g; s/href="//' | 
-        awk -F'[_\.]' '{print $3"."$4"."$5" "$0}' | sort -t. -k1,1n -k2,2n -k3,3n)
+    local links_and_versions="$(curl -s "$url" | grep -o 'href="[^"]*'"$MODEL"'[^"]*\.zip' | sed 's/amp;//g; s/href="//' | 
+        awk -F'[_\.]' '{print $3"."$4"."$5" "$0}' | sort -t. -k1,1n -k2,2n -k3,3n)"
+
+    if [ -z "$links_and_versions" ]
```
Firetop, line 154: I like the error handling, approved.

MerlinAutoUpdate.sh
```diff
@@ -128,26 +134,40 @@ Update_Custom_Settings() {
...
 
-    local latest=$(echo "$links_and_versions" | tail -n 1)
-    local version=$(echo "$latest" | cut -d' ' -f1 | awk -F. '{print $(NF-1)"."$NF}')
-    local link=$(echo "$latest" | cut -d' ' -f2-)
+    local latest="$(echo "$links_and_versions" | tail -n 1)"
+    local linkStr="$(echo "$latest" | cut -d' ' -f2-)"
+    local fileStr="$(echo "$linkStr" | grep -oE "/${MODEL}_[0-9]+.*.zip$")"
+    local versionStr
+
+    if [ -z "$fileStr" ]
```
Firetop, line 162: Will quickly test these changes, but i like how it seems to The changes seem to add flexibility to how versions can be derived. This may be necessary if the format or naming convention of the firmware files changes. However, it also introduces more complexity. The decision to modify this part of the script could be based on observed changes in firmware naming conventions or a need to support more varied names.

MerlinAutoUpdate.sh
```diff
@@ -284,22 +304,57 @@ else
   echo "Cron job 'MerlinAutoUpdate' already exists."
 fi
 
+##-------------------------------------##
+## Added by Martinski W. [2023-Oct-06] ##
+##-------------------------------------##
```
Firetop, line 309: I now follow the reason for the above changes at lines 132. Approved.

Firetop (approved):

Firetop (approved):

MerlinAutoUpdate.sh
Martinski4GitHub: For debugging purposes, I'd suggest creating a separate "Debug Mode" logfile either on the USB-attached disk drive (if available) or in a temporary folder like "/tmp/var/tmp/" so that you can later review it if/when things go sideways. Having the option to generate a more verbose "Debug Mode" logfile can also be extremely useful when the script is "out in the field" and some user is having problems with the initial setup or the actual F/W update process. In such cases, you can instruct them to turn ON "Debug Mode" and supply the resulting debug log. This can save hours trying to diagnose a problem occurring on a user's specific router environment.

My 2 cents.

Firetop: > For debugging purposes, I'd suggest creating a separate "Debug Mode" logfile either on the USB-attached disk drive (if available) or in a temporary folder like "/tmp/var/tmp/" so that you can later review it if/when things go sideways. Having the option to generate a more verbose "Debug Mode" logfile can also be extremely useful when the script is "out in the field" and some user is having problems with the initial setup or the actual F/W update process. In such cases, you can instruct them to turn ON "Debug Mode" and supply the resulting debug log. This can save hours trying to diagnose a problem occurring on a user's specific router environment.
> 
> My 2 cents.

I really like this idea, it's been noted. 
Thank you again for the help btw, while I have scripting experience this is my first time for a merlin router so the guidance on different values to check for, and seeing examples goes a long way!

PR #9, opened 2023-10-10
MerlinAutoUpdate.sh modifications.

Martinski4GitHub:
Modified code to set back LEDs to the same initial state they were before the script was run.
Modified code to prevent saving Username and Password credentials when they're empty.
Other miscellaneous changes.

### DIFF
--- a/MerlinAutoUpdate.sh
+++ b/MerlinAutoUpdate.sh
@@ -216,6 +216,9 @@ Get_Custom_Setting()
     fi
 }
 
+##----------------------------------------##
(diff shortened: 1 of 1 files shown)


Firetop (approved):

Firetop (approved):

Firetop:
Create idea, and as always very nice cleanup.
Makes even the reading of the script a bit easier at places I find lol

Firetop:
Not sure why the heck I opted to set them back to on at the end, now that I think about it, I guess I was just too focused on trying to make it work on my router that once it did I was like "yay! done!" lol.

PR #12, opened 2023-10-13
Update MerlinAutoUpdate.sh

Martinski4GitHub:
Code modifications/additions to use the router's built-in "F/W Update Check" & NVRAM variables to get the latest release update version to set or reset the delayed days for actual updates. Also added a menu item to enable/disable the built-in "F/W Update Check" so that users can control the Auto-Update feature from the same place as the router. Other miscellaneous code improvements.

### DIFF
--- a/MerlinAutoUpdate.sh
+++ b/MerlinAutoUpdate.sh
@@ -7,16 +7,20 @@
 ################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.9"
+readonly SCRIPT_VERSION="0.2.10"
 readonly URL_BASE="https://sourceforge.net/projects/asuswrt-merlin/files"
 readonly URL_RELEASE_SUFFIX="Release"
 
-##-------------------------------------##
-## Added by Martinski W. [2023-Oct-09] ##
-##-------------------------------------##
+##----------------------------------------------##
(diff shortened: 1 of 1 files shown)


Firetop (approved):

Firetop:
So much good work here, I just quickly removed the duplicate "Current firmware version '388.4.0' is up to date." message, otherwise I haven't found any issues with your suggested changes on my router :D 

Many thanks again Martin!

Martinski4GitHub:
> ... I just quickly removed the duplicate "Current firmware version '388.4.0' is up to date." message, ...
> 
Good catch! Honestly, more testing is actually needed under various conditions and on different router models to validate the code. I can only do some limited testing on my own, but I do some unit testing as needed and as my free time permits.

Firetop:
> > ... I just quickly removed the duplicate "Current firmware version '388.4.0' is up to date." message, ...
> 
> Good catch! Honestly, more testing is actually needed under various conditions and on different router models to validate the code. I can only do some limited testing on my own, but I do some unit testing as needed and as my free time permits.

Same with me, I can't always actually test the flash, lots of the time I comment it out, but when time permits I test it all the way through on my unit as well. :D

Thanks so much for your interest in making this as solid as possible, I think it's already come a long way.

PR #17, opened 2023-11-22
Updated MerlinAutoUpdate.sh

Martinski4GitHub:
Some more code improvements & a fix.

- Fixed a problem where a very long base64-encoded string of the login credentials was split into 2 separate lines when generated, so the saved string was split as well which made the login fail every time because of the truncated string. This happened when the length of the raw, original string ("username:password") was greater than 48 chars.

- Improved cleanup & handling of the router LEDs reset after a failure happens during logging in, downloading, uncompressing, or updating the F/W.

- Added code to check if the router login credentials have been saved before starting to download the F/W ZIP file. 

- Added code to prevent running the script multiple instances simultaneously. Only one instance at a time should be allowed to run.

- Added some relevant information at the top of the Main Menu so the user is aware of the router status WRT the current F/W installed vs. any new update versions.

### DIFF
--- a/MerlinAutoUpdate.sh
+++ b/MerlinAutoUpdate.sh
@@ -3,11 +3,11 @@
 # MerlinAutoUpdate.sh
 #
 # Creation Date: 2023-Oct-01 by @ExtremeFiretop.
-# Last Modified: 2023-Nov-20
+# Last Modified: 2023-Nov-21
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.15"
+readonly SCRIPT_VERSION="0.2.16"
 readonly URL_BASE="https://sourceforge.net/projects/asuswrt-merlin/files"
 readonly URL_RELEASE_SUFFIX="Release"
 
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
My friend was curious to try the script and so found/reported a couple of problems:

1) His login credentials were encoded & saved as 2 separate lines so when getting the encoded string from the configuration file it was truncated and, therefore, the login process failed every single time.
[Yeah, he uses very long usernames & passwords - passphrases actually - for almost everything so he stores them all in a password manager app. He tends to be very security-conscious].

2) He noticed he was able to launch 2 separate instances of the script, one per SSH terminal window.

Firetop (approved):

Firetop:
Interesting, my credentials were still saved as a single line, either my creds were short enough, or I should reset everything to zero to properly test the changes.

Merged!

Firetop:
Actually refreshing myself with our to-do list in the readme.

If we are happy with the state of the update functions, naturally the backup solution seems like the next thing to focus on...
But am I nuts to think we would just be re-doing lots of the work that already exists?

Script tools like: ViktorJp/BACKUPMON?
I almost feel like for now we just let people use something like Backupmon, or try and get ViktorJp to contibute some of his knowledge on his backup script to integrate with ours.

Martinski4GitHub:
> Interesting, my credentials were still saved as a single line, either my creds were short enough, ...
> 
Yeah, their concatenation was most likely shorter than 48 chars. The length of "username + password" must be at least **48** chars to have triggered the problem. Like I said, my friend really likes long passphrases :>).

Martinski4GitHub:
> 
> If we are happy with the state of the update functions, naturally the backup solution seems like the next thing to focus on... But am I nuts to think we would just be re-doing lots of the work that already exists?
> 
> Script tools like: ViktorJp/BACKUPMON? I almost feel like for now we just let people use something like Backupmon, or try and get ViktorJp to contibute some of his knowledge on his backup script to integrate with ours.
>
Yes, I think there's no need to "reinvent the wheel" with a separate backup solution. BACKUPMON does the job (I've contributed a little bit to that project as well) and that's what I'd recommend.

Martinski4GitHub:
One thing we were talking about with my friend (as we were bouncing ideas off each other) is that it would be a good idea to have a separate verbose log file for each run of the actual F/W Update process, so there's a record of it when running as a cron job. This would help if something goes wrong as we can review the log entries.

The log filename would be something like "${ModelID}_FW_UPdate_${DATE}_${TIME}.log" (e.g. RT-AX86U_FW_UPdate_2023-11-22_03_00_00.log) and can be saved on a specified directory on a USB-attached drive.

Firetop:
> it would be a good idea to have a separate verbose log file for each run of the actual F/W Update process, so there's a record of it when running as a cron job. This would help if something goes wrong as we can review the log entries.
> 
> The log filename would be something like "${ModelID}_FW_UPdate_${DATE}_${TIME}.log" (e.g. RT-AX86U_FW_UPdate_2023-11-22_03_00_00.log) and can be saved on a specified directory on a USB-attached drive.

It's a great idea, your friend sure he doesn't want to contribute himself? Lol.
Scripting isn't that hard once you sit down and poke ;)

As for the logging the update process, I've taken a stab at it and started to add options for logging control.

Take a look at: https://github.com/Firetop/MerlinAutoUpdate-Router/commit/9e060c50de95e8bef4b63d539b82e310f4970411

Oh and yes, I did make you an official author :)

Firetop:
Oh darn, I forgot to include some kinda log rotation.
Should it be based on amount of logs or amount of days past?

We could do:

```
# Log rotation - delete logs older than 30 days
find "$LOG_BASE_DIR" -name '*.log' -type f -mtime +30 -exec rm {} \;
```

Firetop:
@Martinski4GitHub 

Fixed it just now.

https://github.com/Firetop/MerlinAutoUpdate-Router/commit/ccdf3468e01299d81963fd3bfa4f0f81b484b441

Now it properly logs the output to the desired location, and determines the default location based on if a USB is plugged in or not at the time of install.

Also properly finds and rotates any logs on the USB older than 30 days. I started with 30 days because I have no idea how long peoples delay cycles may be.

Firetop:
On second thought, I don't think users need to be able to configure the log location.
Any use-case you can think? 

Else we just hard code a location and tell them to find it if they need it.

Martinski4GitHub:
> 
> It's a great idea, your friend sure he doesn't want to contribute himself? Lol. Scripting isn't that hard once you sit down and poke ;)
> 
He said he wouldn't feel comfortable touching someone else's code given his limited experience with coding. He's more of a hardware guy.

> As for the logging the update process, I've taken a stab at it and started to add options for logging control.
> 
Yeah, looks like a good solid start.

> 
> Oh and yes, I did make you an official author :)
>
Well, I appreciate that. Thanks.

Martinski4GitHub:
> On second thought, I don't think users need to be able to configure the log location. Any use-case you can think?
> 
> Else we just hard code a location and tell them to find it if they need it.
>
I think it's important to give users the choice of selecting the location for the log files. For example, a USB-attached drive can have multiple partitions (say, one for Entware and a separate partition for "data"). Some users may want to select one or the other partition based on personal preference to store the log files.

Martinski4GitHub:
> 
> ```
> # Log rotation - delete logs older than 30 days
> find "$LOG_BASE_DIR" -name '*.log' -type f -mtime +30 -exec rm {} \;
> ```
30 days sounds good right now. I guess we'll have to see how it works out in real use cases and adjust it.

PR #22, opened 2023-12-03
Update MerlinAU.sh in Dev

Firetop:
1. Changed the ordering of the Menu so more of the advanced stuff is at the bottom and more common stuff at the top.

2. Modified the failure warning to include an extra reason why it may have failed to login.

3. Added a check to make sure the expected mount point and the actual mount point match for USB, else if stops the script.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,11 +4,11 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2021-Nov-01
-# Last Modified: 2023-Dec-01
+# Last Modified: 2023-Dec-03
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.25"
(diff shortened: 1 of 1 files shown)


Firetop:
@Martinski4GitHub 

Is this something I should be doing? Can you approve/merge this for me as a test to make sure you have access?

Thanks again! If you do, then as discussed I'll start doing requests for review as well as we get more mature in the development process.

Martinski4GitHub:
> 
> Radio silence for a while is fine, I may do the same, as long as stuff is documented with a pull request for when the other comes back so we can still review the changes?
> 
> What I mean is if it's small stuff I would probably wait the while, if it was urgent bug that needed fixing and I wasn't here or you weren't here, I expect us to just push it through, but still document for the other to check when they return! :)
> 
That sounds like a good policy: Small stuff can wait. Urgent bug fixes can be pushed & merged. And document additions & changes in the PRs (or the source code itself as appropriate) so they can be reviewed later on.

PR #23, opened 2023-12-04
Update MerlinAU.sh

Firetop:
The script had a bug where it always downloaded to the home directory even if a USB was connected and selected as the destination. This has been resolved at lines: !1502

Also had to modify the reboot check to follow the user selection at lines: ~921

Also fixed a bug where even though I have a rog build in the directory after unzipping, it doesn't prompt me about which build to use at lines: ~1609 and ~1618

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -8,7 +8,7 @@
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.26"
+readonly SCRIPT_VERSION="0.2.27"
 readonly SCRIPT_NAME="MerlinAU"
 
 ##-------------------------------------##
@@ -1518,8 +1518,8 @@ _RunFirmwareUpdateNow_()
     if [ "$FW_ZIP_DIR" != "$FW_BIN_DIR" ] && ! _CreateDirectory_ "$FW_BIN_DIR" ; then return 1 ; fi
 
     # Get current firmware version #
-    ##FOR DEBUG ONLY##current_version="$(_GetCurrentFWInstalledShortVersion_)"	
-    current_version="388.3.0"
+    current_version="$(_GetCurrentFWInstalledShortVersion_)"	
+    ##FOR DEBUG ONLY##current_version="388.3.0"
 
     #---------------------------------------------------------#
     # If the "F/W Update Check" in the WebGUI is disabled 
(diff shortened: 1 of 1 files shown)


Firetop:
That's it for me tonight, if I go any longer the code becomes too sloopy. (It's late for me) lol!

No need to merge this, just wanted you to see what I was working on, there is still some "bugs" to iron out to what I can tell.

Firetop:
> > The script had a bug where it always downloaded to the home directory even if a USB was connected and selected as the destination. This has been resolved at lines: !1502
> 
> Well, I did a quick review of your recent changes, and I think what you meant is that the F/W ZIP file is downloaded correctly to the assigned ZIP directory path; however, when uncompressing the ZIP file, the F/W binary image is always placed in the $HOME directory. Is that what you meant?
> 

Yes your right this is exactly what I mean, that it always used the home directory for the actual unzip and extraction of the zip "post" download.

Now that you explain it, I understand the concern with doing it all from the USB, I didn't think about the fact that it may unmount the USB in the update process. I didn't actually try to flash an update with these latest changes it was just something I was working on to see if I could have it "all happen" on the USB or on the root, so to speak.

But considering the concerns with the system unloading stuff, I think your right and I can probably drop that idea.

The rog build bug still does need to be worked out, but I had only just noticed that an hour or so ago lol. Not anywhere near ready yet which is why I didn't expect you to merge this I just wanted you to kinda see the through process I had.

Thanks for your review and logical thoughts on the matter!

Firetop:
Closed for now, the rog stuff will be it's own independent pull request

Firetop:
It also was an old USB probably many many years old, cause it was only a 4GB USB. 
I don't remember the last time I was able to buy a 4GB USB as a standard size lol. Now like 32Gb is the standard size if not more.

it could of very well have been on it's last legs anyways, and that just pushed it over the edge.

Martinski4GitHub:
> It also was an old USB probably many many years old, cause it was only a 4GB USB. I don't remember the last time I was able to buy a 4GB USB as a standard size lol. Now like 32Gb is the standard size if not more.
> 
> it could of very well have been on it's last legs anyways, and that just pushed it over the edge. RIP USB. Your sacrifice was not in vain. I learned something lol!
>
Yeah, I'd say that's the most likely explanation. Using the standard F/W update process from the webGUI, I've noticed my USB-attached drive getting unmounted almost right away, but I never had the drive corrupted or become unreadable; but then again, the F/W image was not placed on the USB drive itself.

BTW, I should note that I have Entware installed on my USB drive with 2 services running from it, so when the F/W update process starts up, the "/jffs/scripts/unmount" script is executed almost right away and those Entware services are terminated and the drive gets unmounted. Now, if you have just a simple "clean" USB drive (i.e. no Entware installed, no extra scripts/services running from it) and no other services accessing the drive filesystem at the time (SMB, FTP, NFS, DLNA, a logger updating log files, etc.), it's possible that the USB drive may not get unmounted right away, but ultimately it will get unmounted at some point. Whether that's before the F/W image is loaded into RAM and no temporary file is created on the same location where the image was uploaded is another question.

PR #26, opened 2023-12-07
Update MerlinAU.sh

Martinski4GitHub:
Fixed bug causing the ZIP file to be deleted even when located on a USB-attached drive where Entware is installed & some code improvements in "Get_Custom_Setting()" function. Also modified "_DoCleanUp_" function to try to keep the ZIP file from being deleted under some conditions. Minor cleanup.


Firetop:
This looks solid to me. 
I also just tested it, and confirmed the expected behavior on my router.

I think this is pretty much ready for master honestly. 
Any concerns or last changes you'd want to make before pushing this to the "main" branch?

PR #33, opened 2023-12-16
Update MerlinAU.sh

Martinski4GitHub:
Some code improvements WRT getting the currently available RAM that may be "free" to be reclaimed at the time to be able to proceed with the F/W Update process.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,11 +4,11 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2021-Nov-01
-# Last Modified: 2023-Dec-10
+# Last Modified: 2023-Dec-15
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.29"
+readonly SCRIPT_VERSION="0.2.30"
 readonly SCRIPT_NAME="MerlinAU"
 
 ##-------------------------------------##
@@ -941,6 +941,50 @@ _GetCurrentFWInstalledShortVersion_()
     echo "$theVersionStr"
 }
 
+##-------------------------------------##
+## Added by Martinski W. [2023-Dec-15] ##
+##-------------------------------------##
+_HasRouterMoreThan256MBtotalRAM_()
+{
+   local totalRAM_KB
+   totalRAM_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemTotal: /{print $2}')"
+   [ -n "$totalRAM_KB" ] && [ "$totalRAM_KB" -gt 262144 ] && return 0
+   return 1
+}
+
+##-------------------------------------##
+## Added by Martinski W. [2023-Dec-15] ##
+##-------------------------------------##
+#---------------------------------------------------------------------#
+# The actual amount of RAM that is available for any new process
+# (*without* using the swap file) can be roughly estimated from
+# "MemFree" & "Page Cache" (i.e. Active files + Inactive files),
+# This estimate must take into account that the overall system 
+# (kernel + native services + tmpfs) needs a minimum amount of RAM
+# to continue to work, and that not all reclaimable Page Cache can
+# be reclaimed because some may actually be in used at the time.
+#---------------------------------------------------------------------#
+_GetAvailableRAM_KB_()
+{
+   local theMemAvailable_KB  theMemFree_KB
+   local activeFiles_KB  inactiveFiles_KB  thePageCache_KB
+
+   theMemAvailable_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemAvailable: /{print $2}')"
+   [ -n "$theMemAvailable_KB" ] && echo "$theMemAvailable_KB" && return 0
+
+   theMemFree_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemFree: /{print $2}')"
+   activeFiles_KB="$(cat /proc/meminfo | awk -F ' ' '/^Active\(file\): /{print $2}')"
+   inactiveFiles_KB="$(cat /proc/meminfo | awk -F ' ' '/^Inactive\(file\): /{print $2}')"
+   thePageCache_KB="$((activeFiles_KB + inactiveFiles_KB))"
+
+   #----------------------------------------------------------------#
+   # Since not all Page Cache is guaranteed to be reclaimed at any
+   # moment, we simply estimate that only half will be reclaimable.
+   #----------------------------------------------------------------#
+   theMemAvailable_KB="$((theMemFree_KB + ((thePageCache_KB / 2))))"
+   echo "$theMemAvailable_KB" ; return 0
+}
+
 get_free_ram() {
     # Using awk to sum up the 'free', 'buffers', and 'cached' columns.
     free | awk '/^Mem:/{print $4}'  # This will return the available memory in kilobytes.
@@ -973,18 +1017,18 @@ get_required_space() {
 ##------------------------------------------##
 check_memory_and_prompt_reboot() {
     local required_space_kb="$1"
-    local free_ram_kb="$2"
+    local availableRAM_kb="$2"
 
-    if [ "$free_ram_kb" -lt "$required_space_kb" ]; then
+    if [ "$availableRAM_kb" -lt "$required_space_kb" ]; then
         Say "Insufficient RAM available."
 
         # Attempt to clear PageCache #
         Say "Attempting to free up memory..."
         sync; echo 1 > /proc/sys/vm/drop_caches
 
-        # Check memory again #
-        free_ram_kb=$(free | awk '/^Mem:/{print $4 }')
-        if [ "$free_ram_kb" -lt "$required_space_kb" ]; then
+        # Check available memory again #
+        availableRAM_kb=$(_GetAvailableRAM_KB_)
+        if [ "$availableRAM_kb" -lt "$required_space_kb" ]; then
             # During an interactive shell session, ask user to confirm reboot #
             if _WaitForYESorNO_ "Reboot router now"; then
                 _AddPostRebootRunScriptHook_
@@ -993,7 +1037,7 @@ check_memory_and_prompt_reboot() {
                 exit 1  # Although the reboot command should end the script, it's good practice to exit after.
             fi
         else
-            Say "Successfully freed up memory. Available: ${free_ram_kb}KB."
+            Say "Successfully freed up memory. Available: ${availableRAM_kb}KB."
         fi
     fi
 }
@@ -1169,7 +1213,7 @@ change_build_type() {
 
     printf "Current Build Type: ${GRNct}$display_choice${NOct}.\n"
     printf "Would you like to use the original ${REDct}ROG${NOct} themed user interface?${NOct}\n"
-	
+
 	while true; do
 		printf "\n [${theExitStr}] Enter your choice (y/n): "
 		read -r choice
@@ -1626,7 +1670,7 @@ _RunFirmwareUpdateNow_()
     set -- $(_GetLatestFWUpdateVersionFromWebsite_ "$FW_URL_RELEASE")
     release_version="$1"
     release_link="$2"
-	
+
     # Extracting the first octet to use in the curl
     firstOctet="$(echo "$release_version" | cut -d'.' -f1)"
     # Inserting dots between each number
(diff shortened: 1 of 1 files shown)


ExtremeFiretop (approved):

ExtremeFiretop:
All looks great to me!

I created a new PR for the discussed warning messages:

https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/34/files

ExtremeFiretop:
@Martinski4GitHub 

Why is it that since we stopped using get_free_ram() function that my memory in the script no longer follows the memory in the WebUI?

For example, at the backup stage my WebUI got to 55MB of free memory, but the script was still reporting some 330MB of free memory. When I got back to get_free_ram function, now all of a sudden the script follows the WebUI perfectly.

I understand your method is technically more comprehensive, but I feel like it would throw users off actually comparing and monitoring like I just did.

ExtremeFiretop:
And I should mention that while the /proc/meminfo is technically more accurate, it's only so by 2MB to what I can tell lol.
Is it worth the potential confusion?

Martinski4GitHub:
> @Martinski4GitHub
> 
> Why is it that since we stopped using get_free_ram() function that my memory in the script no longer follows the memory in the WebUI?
> 
> For example, at the backup stage my WebUI got to 55MB of free memory, but the script was still reporting some 330MB of free memory. When I got back to get_free_ram function, now all of a sudden the script follows the WebUI perfectly.
> 
> I understand your method is technically more comprehensive, but I feel like it would throw users off actually comparing and monitoring like I just did.
>
From the perspective of the kernel & the OS memory management, the definition of "**free**" memory is different from the definition of "**available**" memory. The webGUI shows users the amount of "**free**" RAM, but what we want to know in the script is the amount of "**available**" RAM because that's more accurate when deciding whether we have enough to proceed with flashing the F/W image.

Some users may be confused but that's likely because of a lack of knowledge & understanding of the difference between "**free**" & "**available**" RAM. Note that the message output by the script is correct & clear:

It says:
**Required RAM**: XXXX KB - **_Available_ RAM**: YYYY KB"

It does ****not**** say:
**Required RAM**: XXXX KB - **_Free_ RAM**: YYYY KB"

Here's a screenshot showing the current memory on my router. Notice where the output from "**/proc/meminfo**" clearly labels the amounts for "**Free**" vs. "**Available**" RAM. Those are the correct amounts, regardless of some users being confused by them.

RT-AC86U_Mem_Free_vs_Available

ExtremeFiretop:
> > @Martinski4GitHub
> > Why is it that since we stopped using get_free_ram() function that my memory in the script no longer follows the memory in the WebUI?
> > For example, at the backup stage my WebUI got to 55MB of free memory, but the script was still reporting some 330MB of free memory. When I got back to get_free_ram function, now all of a sudden the script follows the WebUI perfectly.
> > I understand your method is technically more comprehensive, but I feel like it would throw users off actually comparing and monitoring like I just did.
> 
> From the perspective of the kernel & the OS memory management, the definition of "**free**" memory is different from the definition of "**available**" memory. The webGUI shows users the amount of "**free**" RAM, but what we want to know in the script is the amount of "**available**" RAM because that's more accurate when deciding whether we have enough to proceed with flashing the F/W image.
> 
> Some users may be confused but that's likely because of a lack of knowledge & understanding of the difference between "**free**" & "**available**" RAM. Note that the message output by the script is correct & clear:
> 
> It says: **Required RAM**: XXXX KB - **_Available_ RAM**: YYYY KB"
> 
> It does **not** say: **Required RAM**: XXXX KB - **_Free_ RAM**: YYYY KB"
> 
> Here's a screenshot showing the current memory on my router. Notice where the output from "**/proc/meminfo**" clearly labels the amounts for "**Free**" vs. "**Available**" RAM. Those are the correct amounts, regardless of some users being confused by them.
> 
> RT-AC86U_Mem_Free_vs_Available

Thanks for the great answer!

I'm willing to accept that answer, I understand that it's better for the script to have a more comprehensive view, my concern was really on user interaction.

In this case, in willing to just educate any user that may be confused or claim it isn't working correctly.

I figured maybe if we displayed the actual free ram, not available, then maybe we could avoid this.

Martinski4GitHub:
> 
> In this case, in willing to just educate any user that may be confused or claim it isn't working correctly.
> 
> I figured maybe if we displayed the actual free ram, not available, then maybe we could avoid this.
>
We could easily change all the output messages to display all **3** amounts; something like this:

**Required RAM**: XXXX KB - **_Available_ RAM**: YYYY KB" -  **_Free_ RAM**: ZZZZ KB.

This way, it becomes clearer for users who are non-technical or don't know any better, and they will see the difference between "**Free**" & "**Available**" RAM.

ExtremeFiretop:
> > In this case, in willing to just educate any user that may be confused or claim it isn't working correctly.
> > I figured maybe if we displayed the actual free ram, not available, then maybe we could avoid this.
> 
> We could easily change all the output messages to display all **3** amounts; something like this:
> 
> **Required RAM**: XXXX KB - **_Available_ RAM**: YYYY KB" - **_Free_ RAM**: ZZZZ KB.
> 
> This way, it becomes clearer for users who are non-technical or don't know any better, and they will see the difference between "**Free**" & "**Available**" RAM.

Your right. this is the real solution.
Display free, but let the script actually care about availability. That way users don't freak out and the script still has the more comprehensive view.

ExtremeFiretop:
> > In this case, in willing to just educate any user that may be confused or claim it isn't working correctly.
> > I figured maybe if we displayed the actual free ram, not available, then maybe we could avoid this.
> 
> We could easily change all the output messages to display all **3** amounts; something like this:
> 
> **Required RAM**: XXXX KB - **_Available_ RAM**: YYYY KB" - **_Free_ RAM**: ZZZZ KB.
> 
> This way, it becomes clearer for users who are non-technical or don't know any better, and they will see the difference between "**Free**" & "**Available**" RAM.

Let me fix these since I'm the one that changed them on the PR. I'll submit another shortly

Martinski4GitHub:
BTW, I'm starting to read through the DMs WRT Backupmon; but man, there's a boatload of them!!
I'm not sure I'll read them all tonight. Today was a long day for me, and reading emails/DMs is not really my idea of "relaxing after work." I'll try to get the gist of what you guys discussed & resolved if I can tonight.

ExtremeFiretop:
> BTW, I'm starting to read through the DMs WRT Backupmon; but man, there's a boatload of them!! I'm not sure I'll read them all tonight. Today was a long day for me, and reading emails/DMs is not really my idea of "relaxing after work." I'll try to get the gist of what you guys discussed & resolved if I can tonight.

I didn't even expect you to reply tonight at all!
But I'm kinda happy you did, I'll handle the fixes you found and resubmit for you tomorrow if you have a few seconds.

As for the forums, um yeah... BACKUPMON is acting fishy. Viktor was hoping you could provide some advise because essentially it always unmounts my USB and not only that, also creates mount points which then cause the issue we had previously on our script of mounting as SDA(1) instead of SDA.

Viktor believes it's related to the labels, it's possible, but based on my tests, it's unlikely.

Martinski4GitHub:
> 
> I didn't even expect you to reply tonight at all! But I'm kinda happy you did, I'll handle the fixes you found and resubmit for you tomorrow if you have a few seconds.
> 
> As for the forums, um yeah... BACKUPMON is acting fishy. Viktor was hoping you could provide some advise because essentially it always unmounts my USB and not only that, also creates mount points which then cause the issue we had previously on our script of mounting as SDA(1) instead of SDA.
> 
> Viktor believes it's related to the labels, it's possible, but based on my tests, it's unlikely.
>
IIRC, at some point Viktor was having problems with accurately detecting all the mount points to display them in a menu for users to be able to select, so I provided a function to do that. Then I vaguely remember him having issues with labels being blank, or getting the same label with a numeric suffix. I don't recall all the details, but I thought he had figured out the solution.

I'll likely need to know more context about the problem and when exactly happens so that I can recreate it if I can, or at least have a better idea of what's going on.

Martinski4GitHub:
Also,  IIRC, Backupmon originally did not make backups to the router USB-attached drives. It required a remote destination (e.g. NAS) so you needed to provide a UNC path. I have not followed very closely about what happens when you select a local USB drive as the destination. That's something that could potentially create problems for users when our script runs unattended via a cron job.

ExtremeFiretop:
> Also, IIRC, Backupmon originally did not make backups to the router USB-attached drives. It required a remote destination (e.g. NAS) so you needed to provide a UNC path. I have not followed very closely about what happens when you select a local USB drive as the destination. That's something that could potentially create problems for users when our script runs unattended via a cron job.

According to Viktor the logic is supposed to allow for it to backup to a USB and not unmount. So as long as that's the case I don't expect too many issues. Backupmon would backup to the NAS or USB, and then our script handles entware and other services and then the flash. 

I guess I potentially issues on a small USB or one low on storage if the logic doesn't cleanup like ours does, but otherwise I would think it's fine, right now it just rips the USB right out from under us so not only our script but other scripts can run into issues with this.

Martinski4GitHub:
> > Also, IIRC, Backupmon originally did not make backups to the router USB-attached drives. It required a remote destination (e.g. NAS) so you needed to provide a UNC path. I have not followed very closely about what happens when you select a local USB drive as the destination. That's something that could potentially create problems for users when our script runs unattended via a cron job.
> 
> According to Viktor the logic is supposed to allow for it to backup to a USB and not unmount. So as long as that's the case I don't expect too many issues. Backupmon would backup to the NAS or USB, and then our script handles entware and other services and then the flash.
> 
> I guess I see potentially issues on a small USB or one low on storage if the logic doesn't cleanup like ours does, but otherwise I would think it's fine, right now the issue is it just rips the USB right out from under us so not only our script but other scripts can run into issues with this.
> 
> That could be a show stopper.
>
If at this point there are still issues with unmounting the USB drive unexpectedly, or with having the wrong labels attached to the mount point, I'd suggest postponing the integration with Backumon to a later release (post-AMTM integration) so that those issues can investigated and resolve without any pressures. Viktor may be able to find a solution with our help, but it sounds like it's not going to happen by tomorrow night unless we get lucky.

ExtremeFiretop:
> > > Also, IIRC, Backupmon originally did not make backups to the router USB-attached drives. It required a remote destination (e.g. NAS) so you needed to provide a UNC path. I have not followed very closely about what happens when you select a local USB drive as the destination. That's something that could potentially create problems for users when our script runs unattended via a cron job.
> > 
> > 
> > According to Viktor the logic is supposed to allow for it to backup to a USB and not unmount. So as long as that's the case I don't expect too many issues. Backupmon would backup to the NAS or USB, and then our script handles entware and other services and then the flash.
> > I guess I see potentially issues on a small USB or one low on storage if the logic doesn't cleanup like ours does, but otherwise I would think it's fine, right now the issue is it just rips the USB right out from under us so not only our script but other scripts can run into issues with this.
> > That could be a show stopper.
> 
> If at this point there are still issues with unmounting the USB drive unexpectedly, or with having the wrong labels attached to the mount point, I'd suggest postponing the integration with Backumon to a later release (post-AMTM integration) so that those issues can investigated and resolve without any pressures. Viktor may be able to find a solution with our help, but it sounds like it's not going to happen by tomorrow night unless we get lucky.

I was thinking the same thing

But thelonelycoder doesn't seem too pressured himse.
I'll give Viktor tomorrow. Otherwise we push an update after the fact with the code integration with Backupmon

Martinski4GitHub:
> > 
> > If at this point there are still issues with unmounting the USB drive unexpectedly, or with having the wrong labels attached to the mount point, I'd suggest postponing the integration with Backumon to a later release (post-AMTM integration) so that those issues can investigated and resolve without any pressures. Viktor may be able to find a solution with our help, but it sounds like it's not going to happen by tomorrow night unless we get lucky.
> 
> I was thinking the same thing
> 
> But thelonelycoder doesn't seem too pressured himse. I'll give Viktor tomorrow. Otherwise we push an update after the fact with the code integration with Backupmon.
>
OK, I'll leave it up to you. IMO, tomorrow we should focus on the things we can control ourselves within our script, like making sure we are handling the Alpha & Beta releases correctly (i.e. **not** updating from a master release to an Alpha or Beta version; and based on user selection, to update from an Alpha/Beta version to the next master release). Also, I saw a possible issue (which I put a comment on) about selecting the default mount point for ZIP & LOG files instead of the user-selected destination. As I said in my comment, we need to make sure we honor the user's selection.

ExtremeFiretop:
> > > If at this point there are still issues with unmounting the USB drive unexpectedly, or with having the wrong labels attached to the mount point, I'd suggest postponing the integration with Backumon to a later release (post-AMTM integration) so that those issues can investigated and resolve without any pressures. Viktor may be able to find a solution with our help, but it sounds like it's not going to happen by tomorrow night unless we get lucky.
> > 
> > 
> > I was thinking the same thing
> > But thelonelycoder doesn't seem too pressured himse. I'll give Viktor tomorrow. Otherwise we push an update after the fact with the code integration with Backupmon.
> 
> OK, I'll leave it up to you. IMO, tomorrow we should focus on the things we can control ourselves within our script, like making sure we are handling the Alpha & Beta releases correctly (i.e. **not** updating from a master release to an Alpha or Beta version; and based on user selection, to update from an Alpha/Beta version to the next master release). Also, I saw a possible issue (which I put a comment on) about selecting the default mount point for ZIP & LOG files instead of the user-selected destination. As I said in my comment, we need to make sure we honor the user's selection.

Agreed.

It's not a deal breaker, but I want to get it done. I keep pushing it off for other things like Backupmon today, but like you said I rather just tighten everything up tomorrow so we can send it off.

I also saw your notes, I fixed the one I'm just trying to wrap my brain around the other lol

PR #44, opened 2023-12-19
Update MerlinAU.sh

Martinski4GitHub:
Added a background child process to simply wait for 6 minutes while the F/W Update is in progress. If after 6 minutes the curl process still exists, the background task kills it; otherwise, it does nothing & returns. This is our "escape hatch" to handle the rare possibility that the curl command never completes the F/W Update. OTOH, if the F/W update completes normally, the background task will simply do nothing and may get terminated if the router reboots itself. I believe 6 minutes should be plenty of time to complete any F/W update, but we can increase it if/when needed.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -2092,7 +2092,7 @@ fi
 _CheckForNewScriptUpdates_
 
 ##----------------------------------------##
-## Modified by Martinski W. [2023-Nov-19] ##
+## Modified by Martinski W. [2023-Dec-18] ##
 ##----------------------------------------##
 FW_UpdateCheckState="$(nvram get firmware_check_enable)"
 [ -z "$FW_UpdateCheckState" ] && FW_UpdateCheckState=0
@@ -2115,6 +2115,16 @@ then
         printf "Cron job '${GRNct}${CRON_JOB_TAG}${NOct}' already exists.\n"
     fi
     _AddCronJobRunScriptHook_
+
+    #-----------------------------------------------------------#
+    # Check if router reports a new F/W update is available.
+    # If yes, modify the notification settings accordingly.
+    #-----------------------------------------------------------#
+    current_version="$(_GetCurrentFWInstalledShortVersion_)"
+    release_version="$(_GetLatestFWUpdateVersionFromRouter_)"
+    [ -n "$current_version" ] && [ -n "$release_version" ] && \
+    _CheckNewUpdateFirmwareNotification_ "$current_version" "$release_version"
+
     _WaitForEnterKey_
 fi
 
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,11 +4,11 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2021-Nov-01
-# Last Modified: 2023-Dec-17
+# Last Modified: 2023-Dec-18
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.33"
+readonly SCRIPT_VERSION="0.2.34"
 readonly SCRIPT_NAME="MerlinAU"
 
 ##-------------------------------------##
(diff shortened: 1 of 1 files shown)


ExtremeFiretop (approved):

ExtremeFiretop:
All looks fantastic. I merged it is, seems simple but gives us an "out" if absolutely required in those rare instances.

All I did was adjust some of the sleep timeouts here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/commit/95cb5718497c1e6fc1454533dbcd87ee8661f264

As you originally described it can be lowered it from 6 minutes to 3 and a half. 
I appreciated the extra diligence and validation to make sure the curl is really frozen or dead. 

But really don't expect any router to take longer than 3 minutes as that is the default amount of time any of the WebUIs offers before a saying to manually reboot.

I gave it an extra 30 seconds of play room over the default 3 minutes, just incase it happens to be going slightly over or something slightly unexpected happens, but any longer than that it is most likely frozen or a broken curl request and should be killed and tried again post-reboot. 

In my mind theres no need to have people waiting the extra 3 minutes on top of the default 3 minutes. :)

Martinski4GitHub:
> 
> But really don't expect any router to take longer than 3 minutes as that is the default amount of time any of the WebUIs offers before a saying to manually reboot.
> 
...
> 
> In my mind theres no need to have people waiting the extra 3 minutes on top of the default 3 minutes. :) One of those options where if we see three's signs wee need more time, we can extend it, but highly unlikely.
>
I tend to be very conservative regarding time estimates, even with my own. For example, I always add 20% to 25% to my project task/deliverable estimates just to give me enough breathing room if something unexpected/unforeseen happens. When it comes to 3rd-party estimates, I may add up to 50% depending on my assessment of how conservative & experienced the "estimator" is & whether his previous estimates were on target. That's the reason I went double with 6 minutes because I don't know how accurate ASUS default time is. Also, I'm assuming that most of the time the F/W Update will be run automatically with nobody watching, so the extra 3 minutes seemed harmless. I'd rather have the extra time to ensure the F/W update is really "stuck" before killing it than end up killing the process as is about to finish and possibly resulting in a "bricked" router.

OTOH, you're right that this is highly unlikely to happen, but I'd like to err on the side of caution. One "bricked router" due to the script would be too many (and would feel very responsible & guilty about it :>(.

Anyway, my preference & suggestion (as a compromise) would be to have a 4-minute sleep before killing the process. I don't think the extra 60 seconds would be too much even if someone is doing it manually. When I've done it via the webGUI in the past, I normally get up and make myself a cup of coffee or grab something cold to drink (summer months) and take a little walk. By the time I come back, it's done. My philosophy is "a watched pot takes much longer to boil" so unless I'm testing or expecting trouble, I just let it do its thing while I take a few minutes of solitude :>).

Just my 2 cents.

ExtremeFiretop:
@Martinski4GitHub 

So I tend to actually watch the firmware updates like a hawk everytime lol. 
I like to feel like I control as many elements as possible. (At least if I'm there watching when the power goes out, I can explain what happened haha!)

Normally the update is done and rebooted before the 3 minute timer is up on the GUI. 
When i did the update for my side gig I did it over RDP remotely and it came back online updated ~3 minutes later which means it had already updated, rebooted and re-established network access within that time frame. 

However thats a small sample of newer and powerful routers, I can't speak for all of them, which is why I'm good with the proposed compromise of 4minutes :)

ExtremeFiretop:
Done here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/commit/56663302bce0b9fd284806a63b4faf4b47e1c367

And here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/commit/a24ea3e59963e8076363cb490be715c5cf1d3fac

Martinski4GitHub:
> @Martinski4GitHub
> 
> So I tend to actually watch the firmware updates like a hawk everytime lol. I like to feel like I control as many elements as possible. (At least if I'm there watching when the power goes out, I can explain what happened haha!)
> 
Yeah, I get it. When doing s/w dev. work that's still my M.O. as well. and used to be like that before in my younger years with personal projects when tinkering with home routers & whatnot. Nowadays, I've relaxed a little and take things a little easier on the homefront with my personal projects, hobbies, etc., and take my short walks while something is compiling/running/executing, just to relax and not worry too much. So I do understand where you're coming from.

PR #47, opened 2023-12-22
Update MerlinAU.sh

Martinski4GitHub:
- Added new method to log messages into a file during the F/W Update process. This replaces the previous method that was causing some issues when trying to unmount the USB-attached drive just before the F/W was updated.

- Added code to show the F/W Update notification date on the menu.

- Moved some functions near the top.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,11 +4,11 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2023-Dec-19
+# Last Modified: 2023-Dec-21
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.34"
+readonly SCRIPT_VERSION="0.2.35"
 readonly SCRIPT_NAME="MerlinAU"
 
 ##-------------------------------------##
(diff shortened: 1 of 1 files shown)


ExtremeFiretop (approved):

ExtremeFiretop:
So thank you for refactoring the logging solution. @Martinski4GitHub 

However as I was concerned, I just tested updating to the latest firmware and post reboot my USB was still: "SDA(1)" as seen below:

image

image

Interestingly, if I unmount "SDA(1)" in the WebUI, I am only left with the SDA folder in my mnt location as seen below:

image

image

Why does the unmount when flashing not do the same? 
I am always left with both SDA and SDA(1) post reboot after the flash.

Either something is still holding on to the logging on the USB, or this is some kinda firmware bug we would be working around.

ExtremeFiretop:
@Martinski4GitHub 

This is some type of firmware bug we discovered, when I have NOTHING on the USB. (completely empty) and do a firmware upgrade in the WebUI (Not using our script). I have the same problem.

It reboots with the USB as "SDA(1)" instead of "SDA". 
Because I didn't use our script for the flash, and did nothing but flash in the WebUI from 388.4 to 388.5 with the USB mounted, this tells me it's a firmware bug and nothing in our script is still holding up the USB. (from our side anyways).

I will try to add my mnt location handling code from yesterday as a way to handle this bug as a potential situation that can arise.

ExtremeFiretop:
So far there are 2 ways I can re-create this issue for me.

1. Plugin a USB and simply start a firmware upgrade.
(Upon reboot, the USB has SDA(1) as the mnt point.)

2. In the SSH terminal. Type in: "umount -lf "/tmp/mnt/SDA"
(Confirm it's unmounted in WebUI, and the mnt point folder still exists in SSH/WinSCP, and then reboot. Upon reboot, the USB is now called SDA(1)

However, if I go to the webUI, and simply click "Safely remove" it successfully unmounts AND deletes the mnt point.

ExtremeFiretop:
And have the same bug doing from 388.5 --> Down to 388.4...

And also 388.5 --> To the newest Alpha 388.6

ExtremeFiretop:
I tried implementing this code from yesterday again:

```
    #------------------------------------------------------------#
    # Restart the WebGUI to make sure nobody else is logged in
    # so that the F/W Update can start without interruptions.
    #------------------------------------------------------------#
    /sbin/service restart_httpd && sleep 5
	
	# Check if USBMountPoint is set
    if [ -z "$USBMountPoint" ]; then
        printf "USB mount point is not set. Skipping unmount."
    else
	
        # Unmount the USB drive
        Say "Unmounting the USB drive at $USBMountPoint..."
        umount -lf "$USBMountPoint"
		
		sleep 10

        # Check if the umount command was successful
        if [ $? -eq 0 ]; then
            Say "USB drive successfully unmounted."

            # Check if the mount point directory exists
            if [ -d "$USBMountPoint" ]; then
                Say "Removing the mount point directory..."
                rm -r "$USBMountPoint"

                # Check if the rm command was successful
                if [ $? -eq 0 ]; then
                    Say "Mount point directory successfully removed."
                else
                    Say "Failed to remove the mount point directory."
                fi
            else
                Say "Mount point directory does not exist."
            fi
        else
            Say "Failed to unmount the USB drive."
        fi
    fi

	sleep 5
```

Sadly the same result, upon reboot I had SDA and SDA(1).
It seems no matter what, even if the mount point is deleted, that the reboot will attempt to mount it twice and cause this issue as long as the USB is plugged in at the time of boot.

ExtremeFiretop:
Whoa.

So I reformatted and renamed the USB. It's now NTFS format instead of FAT and it's name is SDE instead of SDA.
Everything was setup perfectly. Only one mount point for SDE, it appeared mounted correctly in the WebUI.

I then reboot and SDA comes back!!!! Along with SDE. So I have both SDA and SDE post reboot when before the reboot everything was only working alone with SDE.

So seems something is stuck "remembering" to create SDA on bootup.

ExtremeFiretop:
Completely removed our script, and now all of a sudden the problem is gone. Upon reboot I only have SDE now and no SDA.

So it is our script, and not a firmware bug. 
Something in the restart addconjob hook is creating the directory for the mnt point on reboot.

ExtremeFiretop:
Problem is at line 858:

```
if [ ! -d "$FW_LOG_DIR" ]
then
    mkdir -p -m 755 "$FW_LOG_DIR"
else
    # Log rotation - delete logs older than 30 days #
    find "$FW_LOG_DIR" -name '*.log' -mtime +30 -exec rm {} \;
fi
```

This runs everytime, and tries to create the log directory on every reboot, in this case my log directory was on on the USB so on every reboot, it tries to create "MerlinAU/logs" under /tmp/mnt/SDA

ExtremeFiretop:
@Martinski4GitHub 

Opened a new Pull request to solve these bugs once and for all:

https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/48

Auto-merge should be enabled for when you approve I believe :)

Thanks again for your help and happy holidays!
Joel S.

Martinski4GitHub:
> Problem is at line 858:
> 
> ```
> if [ ! -d "$FW_LOG_DIR" ]
> then
>     mkdir -p -m 755 "$FW_LOG_DIR"
> else
>     # Log rotation - delete logs older than 30 days #
>     find "$FW_LOG_DIR" -name '*.log' -mtime +30 -exec rm {} \;
> fi
> ```
> 
> This runs everytime, and tries to create the log directory on every reboot, in this case my log directory was on on the USB so on every reboot, it tries to create "MerlinAU/logs" under /tmp/mnt/SDA
>
Yep, that would do it. Excellent troubleshooting!!

ExtremeFiretop:
> > Problem is at line 858:
> > ```
> > if [ ! -d "$FW_LOG_DIR" ]
> > then
> >     mkdir -p -m 755 "$FW_LOG_DIR"
> > else
> >     # Log rotation - delete logs older than 30 days #
> >     find "$FW_LOG_DIR" -name '*.log' -mtime +30 -exec rm {} \;
> > fi
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This runs everytime, and tries to create the log directory on every reboot, in this case my log directory was on on the USB so on every reboot, it tries to create "MerlinAU/logs" under /tmp/mnt/SDA
> 
> Yep, that would do it. Excellent troubleshooting!!

I kept the log rotation in the same location but just changed to the logic to if the directory exists, so the log rotations, and removed the make directory command.

And then moved the original function which does make the directorg in the set log directory function, it seemed to work but if you know a better place to move it feel free!

Martinski4GitHub:
> > > Problem is at line 858:
> > > ```
> > > if [ ! -d "$FW_LOG_DIR" ]
> > > then
> > >     mkdir -p -m 755 "$FW_LOG_DIR"
> > > else
> > >     # Log rotation - delete logs older than 30 days #
> > >     find "$FW_LOG_DIR" -name '*.log' -mtime +30 -exec rm {} \;
> > > fi
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This runs everytime, and tries to create the log directory on every reboot, in this case my log directory was on on the USB so on every reboot, it tries to create "MerlinAU/logs" under /tmp/mnt/SDA
> > 
> > 
> > Yep, that would do it. Excellent troubleshooting!!
> 
> Thank you!!! I kept the log rotation in the same location but just changed to the logic to if the directory exists, do the log rotations, and removed the make directory command.
> 
> And then moved the original function which does make the directory in the set log directory function, it seemed to work but if you know a better place to move it feel free!
>
Yeah, I saw that. Only checking for log rotation is fine where it was so it's done every time the script runs.
I just made a couple of changes for better locations where to double-check that the LOG directory actually exists. See PR #49.

PR #48, opened 2023-12-22
Update MerlinAU.sh Logging Solution to Stop Duplicating Mount Points

ExtremeFiretop:
Tested and successfully solves my issues as discussed here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/47

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -1929,9 +1929,11 @@ Would you like to use the ROG build? (y/n)${NOct}\n"
         #----------------------------------------------------------#
         # In the rare case that the F/W Update gets "stuck" for
         # some reason & the "curl" cmd never returns, we create 
-        # a background child process that sleeps for 4 minutes 
-        # and then kills the "curl" process if it still exists. 
-        # Otherwise, this child process does nothing & returns.        
+        # a background child process that sleeps for 3 and a half
+        # minutes and then kills the "curl" process if it still 
+        # exists. Otherwise, this child process does nothing
+		# & returns. NORMALLY the Curl returns almost instantly
+		# once the upload is complete.
         #----------------------------------------------------------#
         (
            sleep 240
@@ -1943,10 +1945,10 @@ Would you like to use the ROG build? (y/n)${NOct}\n"
         ) &
         wait $curlPID ; curlPID=0
         #----------------------------------------------------------#
-        # Let's wait for 30 seconds here. If the router does not 
+        # Let's wait for 90 seconds here. If the router does not 
         # reboot by itself after the process returns, do it now.
         #----------------------------------------------------------#
-        _Reset_LEDs_ ; sleep 30
+        _Reset_LEDs_ ; sleep 90
         /sbin/service reboot
     else
         Say "${REDct}**ERROR**${NOct}: Login failed. Please try the following:
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -1936,7 +1936,7 @@ Would you like to use the ROG build? (y/n)${NOct}\n"
         # once the upload is complete.
         #----------------------------------------------------------#
         (
-           sleep 240
+           sleep 210
            if [ "$curlPID" -gt 0 ]
            then
                kill -EXIT $curlPID 2>/dev/null || return
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,11 +4,11 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2023-Dec-21
+# Last Modified: 2023-Dec-22
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.35"
+readonly SCRIPT_VERSION="0.2.36"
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
I also adjusted the timeouts for the actual flash/curl.

I know, I hear you saying "WHAT!?!?!?! WE JUST DISCUSSED THIS?!"

But, hear me out, I am actually doing flashes with the curl over and over again in all these tests.
The reason I'm extending the actual reboot timer, and lowering the curl timer, why?

Because 99% of the time, the curl completes instantly, that's not the correct location where we should put longest wait/pause.. The upload is done instantly over it's own "local interface". There is no actual "networking" being done on the "upload" since it's already local. Normally we wait for the upload to complete from our device to the router, but in this case it returns within seconds because it's local already.

The kill wait we implemented is literally only for that 1% of the time when it fails. 
it will timeout and kill at 3 and a half minutes if the curl doesn't instantly return like it does 99% of the time. (Remember, it's instant because we aren't "Uploading" anything to local storage)

The rest of the time (99%) we want the longest wait to be once the curl completes and turns, it "should" reboot itself once updated, but if it doesn't after 90 seconds. We handle it for the router after about 90 seconds once the "curl/upload" is done.

Martinski4GitHub:
> I also adjusted the timeouts for the actual flash/curl.
> 
> I know, I hear you saying "WHAT!?!?!?! WE JUST DISCUSSED THIS?!"
> 
> But, hear me out, I am actually doing flashes with the curl over and over again in all these tests while I know you have only done it once so far. The reason I'm extending the actual reboot timer, and lowering the curl timer, why?
> 
Ah, OK. I get it now. That makes sense and explains why with the previous "sleep 60" the F/W Update would work most of the time since, as you said, the "curl" command itself likely takes less than a second to complete; the rest is the actual flashing of the image *after* shutting down most services, any external user space processes (e.g. Entware services, scripts, etc.) and unmounting any USB-attached drives. So I agree that 90 seconds should be sufficient. Good call.

ExtremeFiretop:
> > I also adjusted the timeouts for the actual flash/curl.
> > I know, I hear you saying "WHAT!?!?!?! WE JUST DISCUSSED THIS?!"
> > But, hear me out, I am actually doing flashes with the curl over and over again in all these tests while I know you have only done it once so far. The reason I'm extending the actual reboot timer, and lowering the curl timer, why?
> 
> Ah, OK. I get it now. That makes sense and explains why with the previous "sleep 60" the F/W Update would work most of the time since, as you said, the "curl" command itself likely takes less than a second to complete; the rest is the actual flashing of the image _after_ shutting down most services, any external user space processes (e.g. Entware services, scripts, etc.) and unmounting any USB-attached drives. So I agree that 90 seconds should be sufficient. Good call.

Bingo! I hope that 90 seconds is long enough. Really what we could do is extend the 90 to 3 minutes like the default. So we 'allow' the curl to take up to 3, but expect it to be done within seconds, and then 'allow" the flash and reboot to take up to 3 minutes, or we reboot it ourselves.

I think that might even be better before we promote to production, but I'll wait on your opinions, your 2 cents are always appreciated haha.

Martinski4GitHub:
> 
> Bingo! I hope that 90 seconds is long enough. Really what we could do is extend the 90 to 3 minutes like the default. So we 'allow' the curl to take up to 3, but expect it to be done within seconds, and then 'allow" the flash and reboot to take up to 3 minutes, or we reboot it ourselves.
> 
I was thinking about that as well while making the changes for PR #49. I fully agree that to be on the conservative side we should increase the 90-sec wait before rebooting to 3 minutes. IOW, both wait times (one for the curl command & one for reboot) should be 3 minutes. My preference is to err on the side of caution.

ExtremeFiretop:
> > Bingo! I hope that 90 seconds is long enough. Really what we could do is extend the 90 to 3 minutes like the default. So we 'allow' the curl to take up to 3, but expect it to be done within seconds, and then 'allow" the flash and reboot to take up to 3 minutes, or we reboot it ourselves.
> 
> I was thinking about that as well while making the changes for PR #49. I fully agree that to be on the conservative side we should increase the 90-sec wait before rebooting to 3 minutes. IOW, both wait times (one for the curl command & one for reboot) should be 3 minutes. My preference is to err on the side of caution.

Do it. 

I think we originally had something like this (6 minutes total) but we compromised, but we compromised the wrong way hahaha! It's okay we caught it early and live and learn. 3 and 3 is the safest bet.

Martinski4GitHub:
> > > Bingo! I hope that 90 seconds is long enough. Really what we could do is extend the 90 to 3 minutes like the default. So we 'allow' the curl to take up to 3, but expect it to be done within seconds, and then 'allow" the flash and reboot to take up to 3 minutes, or we reboot it ourselves.
> > 
> > 
> > I was thinking about that as well while making the changes for PR #49. I fully agree that to be on the conservative side we should increase the 90-sec wait before rebooting to 3 minutes. IOW, both wait times (one for the curl command & one for reboot) should be 3 minutes. My preference is to err on the side of caution.
> 
> Do it.
> 
> I think we originally had something like this (6 minutes total) but we compromised, but we compromised the wrong way hahaha! It's okay we caught it early and live and learn. 3 and 3 is the safest bet.
>
Indeed we did. We had the right idea but the wrong locations.  I'll make the changes now.

PR #50, opened 2023-12-23
Updated MerlinAU.sh

Martinski4GitHub:
Code changes WRT the blinking LEDs. Instead of stopping the LEDs while waiting for a reboot, we restart the LEDs with a slower blinking rate to let users know that some activity is still going on but not the same as before when executing the curl command.


ExtremeFiretop (approved):

MerlinAU.sh
```diff
@@ -1756,12 +1764,13 @@ _RunFirmwareUpdateNow_()
     Say "Required RAM: ${required_space_kb} KB - Available RAM: ${availableRAM_kb} KB"
     check_memory_and_prompt_reboot "$required_space_kb" "$availableRAM_kb"
 
+    trap "_DoCleanUp_ 0 "$keepZIPfile" ; exit 0" HUP INT QUIT TERM
```
ExtremeFiretop, line 1767: Thanks for catching the trap for the cleanup, smart to have it outside the versioning check.

MerlinAU.sh
```diff
@@ -1946,8 +1955,10 @@ Would you like to use the ROG build? (y/n)${NOct}\n"
         #----------------------------------------------------------#
         # Let's wait for 3 minutes here. If the router does not 
         # reboot by itself after the process returns, do it now.
+        # Restart the LEDs with a "slower" blinking rate.
         #----------------------------------------------------------#
-        _Reset_LEDs_ ; sleep 180
+        _Reset_LEDs_ ; Toggle_LEDs 3 & Toggle_LEDs_PID=$!
+        sleep 180 ; _Reset_LEDs_
```
ExtremeFiretop, line 1961: Theres a chance we never reach the last _Reset_LEDs_ with this if the router reboots itself before we order it too at the end of our 180 second timer. Which I think for most models is the case.

MerlinAU.sh
```diff
@@ -155,22 +155,29 @@ Toggle_LEDs()
        Toggle_LEDs_PID=""
        return 1
    fi
+
+   if [ $# -eq 0 ] || [ -z "$1" ] || \
```
ExtremeFiretop, line 159: Very cool idea

ExtremeFiretop:
Merged in the changes, thanks for catching the trap for the cleanup, smart to have it outside the versioning check.

As for the LED blink rate idea, I think it's really cool! 

However; there's a chance we never reach the last:  _Reset_LEDs_ just due to the fact that most of the time once the firmware upgrade is complete it reboots itself, and usually before our 180 timer is done. 

Means I'd have that behavior's again when it's kinda luck/random after the upgrade if your LEDs are on or off

MerlinAU.sh
ExtremeFiretop: Before, the last thing we did to the LEDs before the firmware upgrade started was reset, now it's at the end of the 180 and we may never reach there.

ExtremeFiretop: Could we not utilize this idea, but instead, make it blink slow in the wait cycle while it's postponed and found and update.

So lets say, default is 7 days wait, so for those 7 days, it does a slow cycle as a way to get the users attention an update is available or upcoming. And then when the actual update starts , we cycle the LEDS quickly until we reset at the end like previously?

ExtremeFiretop: I guess this means we would be executing a child process to run for 7 days, right now it just checks one a day to see if your past the 7 days. This adds overhead and complication I guess.

ExtremeFiretop: That being said I just reviewed and noticed that we already start the LEDs as a background child process, which makes sense so the script can continue and focus on other things.

So is there really that much of an impact to have it just run in the background for that longer period of time?

ExtremeFiretop: Alternatively we save the original users LED choice in the settings file before every run of the update, and then reset it from the file in the post reboot hook.

ExtremeFiretop: > Alternatively we save the original users LED choice in the settings file before every run of the update, and then reset it from the file in the post reboot hook.

Create a PR https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/51 for this specifically.

Doesn't address my idea of having the child blink slowly for 7 days, but that seemed more complicated since we would need to probably save the PID, and then kill/manage it 7 days later when we actually run the "firmware update now".

Martinski4GitHub: > Theres a chance we never reach the last _Reset_LEDs_ with this, if the router reboots itself automatically as expected before we tell it too at the end of our 180 second timer. Which I think for most models is the case.
>
The **trap** statement in the script contains a call to "**_DoCleanUp_**" function which calls "**_Reset_LEDs_**" as well so, AFAIK, when the router reboots itself, it terminates all processes with a TERM signal which the trap statement would handle. I have the same mechanism in other scripts and upon regular reboots initiated via "**/sbin/service reboot**" the trap works well. It doesn't work if you just pull the plug or in a power outage.

Now, I'm assuming here that after the F/W Update, the router calls this same kind of reboot but I have not verified this, so I suppose there's a chance that the trap statement may not work if the reboot method is different after a F/W Update.

Martinski4GitHub: > We could utilize this idea but instead, make it blink slow in the wait cycle while it's postponed and found and update.
> 
> So lets say, default is 7 days wait, so for those 7 days, it does a slow cycle as a way to get the users attention an update is available or upcoming. And then when the actual update starts , we cycle the LEDS quickly until we reset at the end like previously so post update the LEDs are back to how they were.
>
I don't like the idea of starting a background task to blink the LEDs for days. First, we would need an easy way for people to stop the blinking LEDs, and logging into the router via SSH to call the script to stop is not simple for non-technical folks (like my in-laws), especially if the schedule checks for new F/W updates at 2am and then "suddenly" the LEDs would start blinking and they would think something is wrong with the router.

Martinski4GitHub: > Alternatively we save the original users LED choice in the settings file before every run of the update, and then reset it from the file in the post reboot hook.
>
I like the idea of saving the initial state of the LEDs in the settings file.

ExtremeFiretop: @Martinski4GitHub 

Wouldn't the variable for _DoCleanUp_ usually be refreshed after the reboot and firmware update?

Lets say we rebooted after the firmware update, but that reboot happened while the LEDs were on instead of my regular off setting..

Well, post reboot the script would now take a value of the LEDs being "On" since that was the last set before the reboot, at the next boot, at this line: readonly LED_InitState="$(nvram get led_disable)" it's now 1 instead of 0 like it was.

Meaning the _DoCleanUp_ and Reset_LEDs will now reset the LEDs to 1 instead of 0, unless we save it into the file, since every load of the script is when we take the readonly LED_InitState="$(nvram get led_disable)

ExtremeFiretop: > > We could utilize this idea but instead, make it blink slow in the wait cycle while it's postponed and found and update.
> > So lets say, default is 7 days wait, so for those 7 days, it does a slow cycle as a way to get the users attention an update is available or upcoming. And then when the actual update starts , we cycle the LEDS quickly until we reset at the end like previously so post update the LEDs are back to how they were.
> 
> I don't like the idea of starting a background task to blink the LEDs for days. First, we would need an easy way for people to stop the blinking LEDs, and logging into the router via SSH to call the script to stop is not simple for non-technical folks (like my in-laws), especially if the schedule checks for new F/W updates at 2am and then "suddenly" the LEDs would start blinking and they would think something is wrong with the router.

No worries, it was just an idea to throw out there. I rather just be sure the LEDs reset post reboot, that was the main concern. here.

Martinski4GitHub:
> 
> However; there's a chance we never reach the last: _Reset_LEDs_ just due to the fact that most of the time once the firmware upgrade is complete it reboots itself, and usually before our 180 timer is done.
> 
> Means I'd have that behavior's again when it's kinda luck/random after the upgrade if your LEDs are on or off. Before, the last thing we did to the LEDs before the firmware upgrade started was reset, now it's at the end of the 180 wait time, and we may never reach there.
>
See my comment above regarding the **trap** statement in the script calling "**DoCleanUp**" function which calls "**Reset_LEDs**" function. Having said that, I'm not against saving the initial LED state in the configuration file to make sure the LEDs are set properly when the "post-reboot" is called.

ExtremeFiretop:
> > However; there's a chance we never reach the last: _Reset_LEDs_ just due to the fact that most of the time once the firmware upgrade is complete it reboots itself, and usually before our 180 timer is done.
> > Means I'd have that behavior's again when it's kinda luck/random after the upgrade if your LEDs are on or off. Before, the last thing we did to the LEDs before the firmware upgrade started was reset, now it's at the end of the 180 wait time, and we may never reach there.
> 
> See my comment above regarding the **trap** statement in the script calling "**DoCleanUp**" function which calls "**Reset_LEDs**" function. Having said that, I'm not against saving the initial LED state in the configuration file to make sure the LEDs are set properly when the "post-reboot" is called.

Check this comment: does this make sense or not really?: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/50#discussion_r1435753923

MerlinAU.sh
Martinski4GitHub: > @Martinski4GitHub
> 
> Wouldn't the variable for _DoCleanUp_ usually be refreshed after the reboot and firmware update?
> 
> Lets say we rebooted after the firmware update, but that reboot happened while the LEDs were on instead of my regular off setting..
> 
> Well, post reboot the script would now take a value of the LEDs being "On" since that was the last set before the reboot, at the next boot, at this line: readonly LED_InitState="$(nvram get led_disable)" it's now 1 instead of 0 like it was.
> 
> Meaning the _DoCleanUp_ and Reset_LEDs will now reset the LEDs to 1 instead of 0, unless we save it into the file, since every load of the script is when we take the readonly LED_InitState="$(nvram get led_disable)
>
The "**_Reset_LEDs_**" function sets the LEDs back to the initial state found when the script was started so if the function is called when the router reboots itself (due to the **trap** statement), the LEDs are back to the initial state after the reboot. But again, this assumes that the router terminates all processes with a TERM signal; if this is not what happens during F/W Updates then all bets are off.

ExtremeFiretop: > > @Martinski4GitHub
> > Wouldn't the variable for _DoCleanUp_ usually be refreshed after the reboot and firmware update?
> > Lets say we rebooted after the firmware update, but that reboot happened while the LEDs were on instead of my regular off setting..
> > Well, post reboot the script would now take a value of the LEDs being "On" since that was the last set before the reboot, at the next boot, at this line: readonly LED_InitState="$(nvram get led_disable)" it's now 1 instead of 0 like it was.
> > Meaning the _DoCleanUp_ and Reset_LEDs will now reset the LEDs to 1 instead of 0, unless we save it into the file, since every load of the script is when we take the readonly LED_InitState="$(nvram get led_disable)
> 
> The "**_Reset_LEDs_**" function sets the LEDs back to the initial state found when the script was started so if the function is called when the router reboots itself (due to the **trap** statement), the LEDs are back to the initial state after the reboot. But again, this assumes that the router terminates all processes with a TERM signal; if this is not what happens during F/W Updates then all bets are off.

I'm just not sure I understand this. 

I'll need more testing and research. For now I'm holding off on merging my change until I understand better. 

If the variable isn't saved and the memory is flushed between reboots I just don't understand how a variable could be passed on from one run of the script to another run of the script

Martinski4GitHub:
> 
> Check this comment: does this make sense or not really?: [#50 (comment)](https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/50#discussion_r1435753923)
>
The **trap** statement working is predicated on the assumption that the reboot terminates all processes with the TERM signal which by definition gives each process a chance to exit "gracefully" by doing whatever cleanup is needed (which includes any **trap** statements). Now, I know this works under "normal" reboots called via "/sbin/service reboot" command; however, if this is not the same mechanism used during the F/W Updates then yes, there would be a problem with the initial LEDs state being potentially different after each reboot initiated by the router right after F/W updates.

ExtremeFiretop:
> > Check this comment: does this make sense or not really?: [#50 (comment)](https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/50#discussion_r1435753923)
> 
> The **trap** statement working is predicated on the assumption that the reboot terminates all processes with the TERM signal which by definition gives each process a chance to exit "gracefully" by doing whatever cleanup is needed (which includes any **trap** statements). Now, I know this works under "normal" reboots called via "/sbin/service reboot" command; however, if this is not the same mechanism used during the F/W Updates then yes, there would be a problem with the initial LEDs state being potentially different after each reboot initiated by the router right after F/W updates.

Is the trap a background process? If I'm understanding correctly. What your saying is we launch this trap, the script moves on, this trap is still active though, and it's waiting for a TERM signal from the router, and if it receives one than it does the cleanup and resets the LEDS before it completely shuts down?

I think that's the part that confuses me. I didn't understand the trap as a background process that remained active as the script moves on.

If it wasn't a background process, then it's too high up in the flash function to actually catch the TERM signal initiated below firmware update/curl.

Martinski4GitHub:
> > > Check this comment: does this make sense or not really?: [#50 (comment)](https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/50#discussion_r1435753923)
> > 
> > 
> > The **trap** statement working is predicated on the assumption that the reboot terminates all processes with the TERM signal which by definition gives each process a chance to exit "gracefully" by doing whatever cleanup is needed (which includes any **trap** statements). Now, I know this works under "normal" reboots called via "/sbin/service reboot" command; however, if this is not the same mechanism used during the F/W Updates then yes, there would be a problem with the initial LEDs state being potentially different after each reboot initiated by the router right after F/W updates.
> 
> Is the trap a background process? If I'm understanding correctly. What your saying is we launch this trap, the script moves on, this trap is still active though, and it's waiting for a TERM signal from the router, and if it receives one than it does the cleanup and resets the LEDS before it completely shuts down?
> 
> I think that's the part that confuses me. I didn't understand the trap as a background process that remained active as the script moves on.
> 
> If it wasn't a background process, then it's too high up in the flash function to actually catch the TERM signal initiated below firmware update/curl.
>
The **trap** statement is not a background function. It's essentially a mechanism that sets "signal traps" in the intertask communications interface so that, for example, when you type "Ctrl-C" on your keyboard while a process is running in the foreground, the SIGINT signal is sent and caught by the trap routine to do whatever "cleanup" is supposed to do before exiting. If no "signal traps" are set by the running process, the signal is never detected and just terminates.

IOW, the **trap** cmd does not actually execute the arguments within the statement when called in the script, it simply sets up the "signal traps" to react when any of the signals specified in the statement is caught via the intertask communication interface.

MerlinAU.sh
Martinski4GitHub: > > > @Martinski4GitHub
> > > Wouldn't the variable for _DoCleanUp_ usually be refreshed after the reboot and firmware update?
> > > Lets say we rebooted after the firmware update, but that reboot happened while the LEDs were on instead of my regular off setting..
> > > Well, post reboot the script would now take a value of the LEDs being "On" since that was the last set before the reboot, at the next boot, at this line: readonly LED_InitState="$(nvram get led_disable)" it's now 1 instead of 0 like it was.
> > > Meaning the _DoCleanUp_ and Reset_LEDs will now reset the LEDs to 1 instead of 0, unless we save it into the file, since every load of the script is when we take the readonly LED_InitState="$(nvram get led_disable)
> > 
> > 
> > The "**_Reset_LEDs_**" function sets the LEDs back to the initial state found when the script was started so if the function is called when the router reboots itself (due to the **trap** statement), the LEDs are back to the initial state after the reboot. But again, this assumes that the router terminates all processes with a TERM signal; if this is not what happens during F/W Updates then all bets are off.
> 
> I'm just not sure I understand this.
> 
> I'll need more testing and research. For now I'm holding off on merging my change until I understand better.
> 
> If the variable isn't saved and the memory is flushed between reboots I just don't understand how a variable could be passed on from one run of the script to another run of the script
>
See my comment about the LEDs state always being saved/stored in NVRAM. IOW, runtime variables are used only during execution. Between reboots and separate executions, NVRAM variables are used to read & write values that need to persist across reboots (i.e. non-volatile).

Martinski4GitHub:
> > > > Check this comment: does this make sense or not really?: [#50 (comment)](https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/50#discussion_r1435753923)
> > > 
> > > 
> > > The **trap** statement working is predicated on the assumption that the reboot terminates all processes with the TERM signal which by definition gives each process a chance to exit "gracefully" by doing whatever cleanup is needed (which includes any **trap** statements). Now, I know this works under "normal" reboots called via "/sbin/service reboot" command; however, if this is not the same mechanism used during the F/W Updates then yes, there would be a problem with the initial LEDs state being potentially different after each reboot initiated by the router right after F/W updates.
> > 
> > 
> > Is the trap a background process? If I'm understanding correctly. What your saying is we launch this trap, the script moves on, this trap is still active though, and it's waiting for a TERM signal from the router, and if it receives one than it does the cleanup and resets the LEDS before it completely shuts down?
> > I think that's the part that confuses me. I didn't understand the trap as a background process that remained active as the script moves on.
> > If it wasn't a background process, then it's too high up in the flash function to actually catch the TERM signal initiated below firmware update/curl.
> 
> The **trap** statement is not a background function. It's essentially a mechanism that sets "signal traps" in the intertask communications interface so that, for example, when you type "Ctrl-C" on your keyboard while a process is running in the foreground, the SIGINT signal is sent and caught by the trap routine to do whatever "cleanup" is supposed to do before exiting. If no "signal traps" are set by the running process, the signal is still detected but acted upon and the process just terminates.
> 
> IOW, the **trap** cmd does not actually execute the arguments within the statement when called in the script, it simply sets up the "signal traps" to react when any of the signals specified in the statement is caught via the intertask communication interface.

MerlinAU.sh
ExtremeFiretop: > > > > @Martinski4GitHub
> > > > Wouldn't the variable for _DoCleanUp_ usually be refreshed after the reboot and firmware update?
> > > > Lets say we rebooted after the firmware update, but that reboot happened while the LEDs were on instead of my regular off setting..
> > > > Well, post reboot the script would now take a value of the LEDs being "On" since that was the last set before the reboot, at the next boot, at this line: readonly LED_InitState="$(nvram get led_disable)" it's now 1 instead of 0 like it was.
> > > > Meaning the _DoCleanUp_ and Reset_LEDs will now reset the LEDs to 1 instead of 0, unless we save it into the file, since every load of the script is when we take the readonly LED_InitState="$(nvram get led_disable)
> > > 
> > > 
> > > The "**_Reset_LEDs_**" function sets the LEDs back to the initial state found when the script was started so if the function is called when the router reboots itself (due to the **trap** statement), the LEDs are back to the initial state after the reboot. But again, this assumes that the router terminates all processes with a TERM signal; if this is not what happens during F/W Updates then all bets are off.
> > 
> > 
> > I'm just not sure I understand this.
> > I'll need more testing and research. For now I'm holding off on merging my change until I understand better.
> > If the variable isn't saved and the memory is flushed between reboots I just don't understand how a variable could be passed on from one run of the script to another run of the script
> 
> See my comment about the LEDs state always being saved/stored in NVRAM. IOW, runtime variables are used only during execution. Between reboots and separate executions, NVRAM variables are used to read & write values that need to persist across reboots (i.e. non-volatile).

I understand better, this was an older comment from an hour ago. 

I think I brought the conversation here:

https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/51#issuecomment-1868423028

PR #51, opened 2023-12-24
Update MerlinAU.sh LED Patch

ExtremeFiretop:
As discussed in your PR: 50 https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/50

The simple solution for now to make sure we reset even if it reboots itself is to save the original perference in the settings file and reset it post-reboot.

Let me know what you think of this solution.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -520,7 +520,6 @@ _Init_Custom_Settings_Config_()
          echo "FW_New_Update_Cron_Job_Schedule=\"${FW_Update_CRON_DefaultSchedule}\""
          echo "FW_New_Update_ZIP_Directory_Path=\"${FW_Update_ZIP_DefaultSetupDIR}\""
          echo "FW_New_Update_LOG_Directory_Path=\"${FW_Update_LOG_BASE_DefaultDIR}\""
-         echo "LED_InitState TBD"
       } > "$SETTINGSFILE"
       return 1
    fi
@@ -830,6 +829,9 @@ _Init_Custom_Settings_Config_
 
 # Get the current LED_InitState from the settings file
 current_LED_InitState=$(Get_Custom_Setting LED_InitState)
+
+echo "$current_LED_InitState"
+sleep 20
 
 # Update the LED_InitState in the settings file only if it's TBD (AKA First Run)
 if [ "$current_LED_InitState" = "TBD" ]; then
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -830,9 +830,6 @@ _Init_Custom_Settings_Config_
 # Get the current LED_InitState from the settings file
 current_LED_InitState=$(Get_Custom_Setting LED_InitState)
 
-echo "$current_LED_InitState"
-sleep 20
-
 # Update the LED_InitState in the settings file only if it's TBD (AKA First Run)
 if [ "$current_LED_InitState" = "TBD" ]; then
     Update_Custom_Settings LED_InitState "$LED_InitState"
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,11 +4,11 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2023-Dec-22
+# Last Modified: 2023-Dec-23
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.38"
+readonly SCRIPT_VERSION="0.2.39"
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
Done for now. Seems to work for me when creating a new settings file or when I already have a settings file, correctly updates both at first run and when running the update firmware now function, else it resets the LEDs to what is in the file in a post-reboot situation.

Martinski4GitHub:
> Done for now. Seems to work for me when creating a new settings file or when I already have a settings file, correctly updates both at first run and when running the update firmware now function, else it resets the LEDs to what is in the file in a post-reboot situation.
>
I'm cool with it to ensure the LEDs are back to the original user-defined setting.

ExtremeFiretop:
> > Done for now. Seems to work for me when creating a new settings file or when I already have a settings file, correctly updates both at first run and when running the update firmware now function, else it resets the LEDs to what is in the file in a post-reboot situation.
> 
> I'm cool with it to ensure the LEDs are back to the original user-defined setting.

I'm going to retest this more thoroughly before merging. 

If you think the trap should save the original between reboots then there's no point in any of the additions I made here.

My understanding was that on the new boot after the firmware update, that is a new run of the script, and it wouldn't remember the readonly variable from the previous run that initiated the reboot, but I could be completely wrong.

Basically I thought all the memory was flushed on the reboot so unless it was saved to a file there would be no way For the script to remember what it was before the reboot.

ExtremeFiretop:
> > Done for now. Seems to work for me when creating a new settings file or when I already have a settings file, correctly updates both at first run and when running the update firmware now function, else it resets the LEDs to what is in the file in a post-reboot situation.
> 
> I'm cool with it to ensure the LEDs are back to the original user-defined setting.

My understanding was wrong, basically it doesn't save the variable between reboots, it just resets the LEDS before the shutdown, if it receives a graceful exit signal. (Even if we didn't get past the 180 wait).

But that wouldn't work if the exit wasn't graceful. 
Like you said, a power outage, or if the firmware update is more aggressive in its reboot and just "pulls the rug out" from our script.

In which case the trap wouldn't execute since it didn't receive the TERM signal.

Martinski4GitHub:
> 
> My understanding was that on the new boot after the firmware update, that is a new run of the script, and it wouldn't remember the readonly variable from the previous run that initiated the reboot, but I could be completely wrong.
> 
> Basically I thought all the memory was flushed on the reboot so unless it was saved to a file there would be no way For the script to remember what it was before the reboot.
>
I think I know where the confusion is. Keep in mind that the value of the initial LEDs state is always found/stored in **NVRAM**, and this is where the temporarily saved runtime variable value is stored back to when calling the "**_Reset_LEDs_**" function (because all runtime variables disappear when the script terminates). Therefore, upon reboot the NVRAM variable would be intact to be read back by the script as the initial LEDs state. 

Now, if the F/W Update requires a reset to factory defaults then even the NVRAM vars are reset so *all* user-defined settings are lost, but this is not the normal case.

ExtremeFiretop:
Is there still value to adding this code or should we let the TERM signal handle it?

> > > Done for now. Seems to work for me when creating a new settings file or when I already have a settings file, correctly updates both at first run and when running the update firmware now function, else it resets the LEDs to what is in the file in a post-reboot situation.
> > 
> > 
> > I'm cool with it to ensure the LEDs are back to the original user-defined setting.
> 
> My understanding was wrong, basically it doesn't save the variable between reboots, it just resets the LEDS before the shutdown, if it receives a graceful exit signal. (Even if we didn't get past the 180 wait).
> 
> But that wouldn't work if the exit wasn't graceful. Like you said, a power outage, or if the firmware update is more aggressive in its reboot and just "pulls the rug out" from our script.
> 
> In which case the trap wouldn't execute since it didn't receive the TERM signal.

Martinski4GitHub:
> 
> But that wouldn't work if the exit wasn't graceful. Like you said, a power outage, or if the firmware update is more aggressive in its reboot and just "pulls the rug out" from our script.
> 
> In which case the trap wouldn't execute since it didn't receive the TERM signal.
>
Yes, that is essentially the crux of the **trap** statement working. It's meant to work under regularly initiated reboots where a SIGTERM, SIGINT, SIGHUP, or SIGQUIT signal (**not** a SIGKILL signal) is used to terminate the process. If the F/W Update indeed uses the SIGKILL signal (for example), the "cleanup" meant to be done via the **trap** statement would not be executed.

So the key now is to test/verify what kind of reboot the F/W update initiates. That would determine where we are WRT the **trap** statement.

ExtremeFiretop:
As discussed in PR https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/52

This is not required. Closing without merge.

PR #56, opened 2023-12-26
Extreme firetop cron/memory patch

ExtremeFiretop:
Resolve issues when running as a cron job

Need to allow a default/completely automated path for the script when running as a cron.


ExtremeFiretop:
Tested successfully. Will push to Dev.

PR #62, opened 2024-01-05
Update MerlinAU.sh Cron Menu to Prompt User

ExtremeFiretop:
When running the script the first time, it force creates a CRON job straight out of the gate. 

I am recommending (as per [Viktor Jaep](https://www.snbforums.com/members/viktor-jaep.53834/) in the forums) asking for user input whether or not someone wants that enabled. 
People are typically skiddish about some script making drastic changes to their system and not knowing how to back out of it. 

Might be nice to just get their approval first. 
And have a note that they can always enable/disable the cron job at any time through the menu.


ExtremeFiretop:
I didn't add anything for an $isInteractive flag. I don't believe it's required because if they select "n" then the auto-update and cron is disabled which means it can only run ineractively, no more background job to run the script anyways.

Feel free to correct me if I am wrong :)

Martinski4GitHub:
> 
> Might be nice to just get their approval first. And have a note that they can always enable/disable the cron job at any time through the menu.
>
Yeah, I agree; that's a good idea.

PR #64, opened 2024-01-05
Update MerlinAU.sh Allowing Unsupported Users To Uninstall Completely

ExtremeFiretop:
Currently with our supported model and minimum version checks, we simply block the script from running flat out if they don't meet requirements right out the gate.

If we were to ever get enough traction to integrate with AMTM, this script would be available to all models (even unsupported models) in the AMTM list.

Therefore, technically there could be a situation where an unsupported model downloads the script and runs it through AMTM but it simply quits right away and kicks them back AMTM. This isn't very user friendly. And leaves our script behind in their jffs folder if they don't know how to clean it up since they can never reach our menu to uninstall.

Instead, I now allow the script to install and run up to the menu, but it prompts that it's unsupported in the main menu, and if the _RunFirmwareUpdateNow_() function is triggered anyways, it simply kicks them back to the main menu of our script after explaining it's unsupported due to x reason. 

The user can then manually action the required firmware update manually or uninstall.

This also allows the script to be installed instead of blocked in a situation where someone does manually update their firmware to reach the minimum firmware version.


ExtremeFiretop:
We could even take it a step further with a "would you like to uninstall now?" If it's an Unsupported Model, And also identify the current minimum we have set in the messages.

ExtremeFiretop:
There, I took it a step further by directly adding the option to uninstall if unsupported and ran here: 

https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/64/commits/2929ed77db5916aa827f1b3b99bd1cb07f48b3d6

Martinski4GitHub:
> 
> Instead, I now allow the script to install and run up to the menu, but it prompts that it's unsupported in the main menu, and if the _RunFirmwareUpdateNow_() function is triggered anyways, it simply kicks them back to the main menu of our script after explaining it's unsupported due to x reason.
> 
> The user can then manually action the required firmware update manually or uninstall.
> 
> This also allows the script to be installed instead of blocked in a situation where someone does manually update their firmware to reach the minimum firmware version.
>
Excellent points & solution. Approved.

Martinski4GitHub:
> There, I took it a step further by directly adding the option to uninstall if unsupported and ran here:
> 
Nice!!

PR #71, opened 2024-01-07
Update MerlinAU.sh

Martinski4GitHub:
Fixed problem with setting of "**isInteractive**" variable which was preventing the cron job from running successfully (since it runs in non-interactive mode).

The issue was that in non-interactive mode the prompt variable (**$PS1**) value could vary between different shell interpreters (e.g. ash, bash, sh, etc.), and in some instances, the variable is still defined even though it's not used at all. So we don't rely on that variable anymore. Also, the return value of the `tty` cmd must be checked since it does return a value even in non-interactive mode (which was unexpected).


MerlinAU.sh
```diff
@@ -1135,14 +1135,13 @@ check_memory_and_prompt_reboot() {
         availableRAM_kb=$(_GetAvailableRAM_KB_)
         if [ "$availableRAM_kb" -lt "$required_space_kb" ]; then
             # In an interactive shell session, ask user to confirm reboot #
-            if [ "$inMenuMode" = true ]; then
```
Martinski4GitHub, line 1138: The reason to use the "$isInteractive" variable along with the prompt function on the same line is that we need to check for **both** conditions as a logical **AND** boolean so that if either one is false we fall back on the "**else**" statement to call the "cleanup & exit" procedures. The way it was before, if the user answers NO, there's no code to do the cleanup & exit.

MerlinAU.sh
```diff
@@ -1987,17 +1986,18 @@ Please manually update to version $minimum_supported_version or higher to use th
...
+    ## Modified by Martinski W. [2024-Jan-06] ##
+    ##----------------------------------------##
     availableRAM_kb=$(_GetAvailableRAM_KB_)
     Say "Required RAM: ${required_space_kb} KB - Available RAM: ${availableRAM_kb} KB"
     check_memory_and_prompt_reboot "$required_space_kb" "$availableRAM_kb"
 
     routerURLstr="$(_GetRouterURL_)"
     Say "Router Web URL is: ${routerURLstr}"
 
-    if [ "$inMenuMode" = true ]; then
```
Martinski4GitHub, line 2000: Here we need to use the "**isInteractive**" variable instead of "**inMenuMode**" because we need the account for the case where the script is called with the "**runnow**" argument during an interactive shell session so it's not always called in "menu mode." IOW, there are 3 variations of the call: 1) via the menu option, 2) via a cron job, and 3) via CLI (i.e. interactive shell).

ExtremeFiretop (approved):

ExtremeFiretop:
Downgrading now. Will be back in a few.

ExtremeFiretop:
Bingo! Worked like a charm, 0.2.51 is now gonna be our new master release and $isInteractive flag will be the new standardfor anymore prompts like this going forwards.

Amazing find Martin! Your a genius!

PR #77, opened 2024-01-21
Update MerlinAU.sh

Martinski4GitHub:
Added code to run a check & see if a new F/W update is available. This is done ***ONLY*** if the user has enabled F/W Update Checks; otherwise, we have to respect the user's selection not to check.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,11 +4,11 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Jan-07
+# Last Modified: 2024-Jan-21
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.52"
+readonly SCRIPT_VERSION="0.2.53"
 readonly SCRIPT_NAME="MerlinAU"
 
 ##-------------------------------------##
@@ -2269,6 +2269,9 @@ then
         printf "Cron job '${GRNct}${CRON_JOB_TAG}${NOct}' already exists.\n"
         _AddCronJobRunScriptHook_
     fi
+
+    # Check if there's a new F/W update available #
+    [ -x "$FW_UpdateCheckScript" ] && sh $FW_UpdateCheckScript 2>&1 &
     _WaitForEnterKey_
 fi
 
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -2230,13 +2230,14 @@ fi
 # to check if there's a new version update to notify the user #
 _CheckForNewScriptUpdates_
 
-##------------------------------------------##
-## Modified by ExtremeFiretop [2024-Jan-04] ##
-##------------------------------------------##
+##----------------------------------------##
+## Modified by Martinski W. [2024-Jan-21] ##
+##----------------------------------------##
 FW_UpdateCheckState="$(nvram get firmware_check_enable)"
 [ -z "$FW_UpdateCheckState" ] && FW_UpdateCheckState=0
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
Crap. I merged directly to the branch instead of making a new one..

Well look at this, while I roll back lol!: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/commit/af3bd4ce5538be25c919d3f47468a9d1aff7f53a

ExtremeFiretop:
Okay. Sorry about that, rolled back the changes to dev, let me make a new branch this time lol!

ExtremeFiretop:
Made a new branch and PR so we can compare properly.

ExtremeFiretop:
Okay so I see, you just threw it in the cron check at the start of the script. While I specifically put it past the disabled check of the run now function.

Difference is yours would run at the start of the script everytime if the update is enabled, while mine would only run if the update is enabled and the user selects to check for updates, not at the start of every script run.

I wanted this to be interactive only since if it was a cron job I would let it rely on the automatic watchdog at 2AM.
However reviewing your suggestion, I kinda like the idea of checking everytime our script runs, basically you end up with having the watchdog that checks at 2AM every morning, and our check which checks whenever the cron is run as configured by the user.

Any thoughts?

Martinski4GitHub:
> Okay so I see, you just threw it in the cron check at the start of the script. While I specifically put it past the disabled check of the run now function.
> 
> Difference is yours would run at the start of the script everytime if the update is enabled, while mine would only run if the update is enabled and the user selects to check for updates, not at the start of every script run.
> 
> I wanted this to be interactive only since if it was a cron job I would let it rely on the automatic watchdog at 2AM. However reviewing your suggestion, I kinda like the idea of checking everytime our script runs, basically you end up with having the watchdog that checks at 2AM every morning, and our check which checks whenever the cron is run. ( as configured by the user.), lets say 12AM by default)
> 
> Any thoughts?
>
Yeah, I put the call at the start of the script to run every time because I thought that it wouldn't hurt to do it based on when the script runs (i.e. manual run or based on the cron schedule) since the user has control over that versus the F/W schedule that runs around 2am.  This way, if there's a new F/W Update the day before, the user doesn't have to wait until the following morning at 2am; instead during a manual run or the the cron job running at, say, 12am would catch thee new F/W version available.

P.S. Just updated PR #77 with an improvement to make sure we honor the user's selection.

ExtremeFiretop:
> > Okay so I see, you just threw it in the cron check at the start of the script. While I specifically put it past the disabled check of the run now function.
> > Difference is yours would run at the start of the script everytime if the update is enabled, while mine would only run if the update is enabled and the user selects to check for updates, not at the start of every script run.
> > I wanted this to be interactive only since if it was a cron job I would let it rely on the automatic watchdog at 2AM. However reviewing your suggestion, I kinda like the idea of checking everytime our script runs, basically you end up with having the watchdog that checks at 2AM every morning, and our check which checks whenever the cron is run. ( as configured by the user.), lets say 12AM by default)
> > Any thoughts?
> 
> Yeah, I put the call at the start of the script to run every time because I thought that it wouldn't hurt to do it based on when the script runs (i.e. manual run or based on the cron schedule) since the user has control over that versus the F/W schedule that runs around 2am. This way, if there's a new F/W Update the day before, the user doesn't have to wait until the following morning at 2am; instead during a manual run or the the cron job running at, say, 12am would catch the new F/W version available.
> 
> P.S. Just updated PR #77 with an improvement to make sure we honor the user's selection.

Agreed.

I like the idea of kinda "adding" a update check the user has control over, since the 2am watchdog the user cannot control.

What I was doing was kinda trying to "play nice" side by side with the 2am watchdog. But this is more control by adding an additional watchdog the user controls .

For an update script that's a better implementation. I'll merge yours in now and adjust my PR.

PR #81, opened 2024-01-22
Dev 0.2.53 Target as Next Master release

ExtremeFiretop:
Dev 0.2.53 Target as Next Master release
Tested successfully and confirmed both our charges today seem to work as expected.

1. Migrate the settings folder from MerlinAU if it exists to MerlinAU.d to follow set standards.
2. If update checks are enabled, run the webs_update.sh script before running ours.


ExtremeFiretop:
The migration function can be removed after 2 release cycles.
Making a note that this code is just temporary to cleanup any existing installs.

ExtremeFiretop:
Enabled Auto-Merge for once @Martinski4GitHub approves.
Gives him a chance to squeeze in any last minute changes or concerns.

Martinski4GitHub:
> The migration function can be removed after 2 release cycles. Making a note that this code is just temporary to cleanup any existing installs.
> 
I don't think it causes any harm leaving that code longer than 2 release cycles. For example, I have read that some people have skipped one release version due to issues reported that affect their home network, so they wait for the next version. Then again, there are probably very few people (single-digit number?) who are now actively using the current "master" Auto Update script for which the migration will be necessary.

ExtremeFiretop:
> people

Agreed. The reason I said 2 was because right now we are single digit users, but your right technically there's no harm id we leave it longer. That being said, I got the post ready for SNB when we push this out also. :)

Martinski4GitHub (approved):
All good to go, IMO.

PR #82, opened 2024-01-22
Update MerlinAU.sh ROG Patch

ExtremeFiretop:
In response to this message: https://www.snbforums.com/threads/introducing-merlinau-the-ultimate-firmware-auto-updater-addon.88577/post-887542

This is the proposed solution, all GT models are ROG models, if the product ID is a GT-Anything, display the ROG config options.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -1907,7 +1907,7 @@ Please manually update to version $minimum_supported_version or higher to use th
     fi
 
     ##---------------------------------------##
-    ## Added by ExtremeFiretop [2024-Jan-06] ##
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
Also made commits to block the following models as per RMerlins strong recommendation related to safety.

Didn't just block the models but also disabled support for the entire trx firmware format.

Martinski4GitHub:
> Also made commits to block the following models as per RMerlins strong recommendation related to safety.
> 
> Didn't just block the models but also disabled support for the entire trx firmware format.
>
I agree. At this point, it would not be worth our time trying to figure out issues with these older AC models.

PR #88, opened 2024-01-25
Update MerlinAU.sh

Martinski4GitHub:
Added code to support email notifications for the following events:

- When the script detects that a new F/W Update version is available.
- At the start of flashing a new F/W Update version.
- At the completion of flashing the F/W - post reboot.

The email notification mechanism uses AMTM email configuration setup so the option for F/W Update notifications becomes visible on the menu only if the user has chosen to enable and set up email notifications via AMTM.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,11 +4,11 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Jan-23
+# Last Modified: 2024-Jan-24
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION="0.2.56"
+readonly SCRIPT_VERSION="0.2.57"
 readonly SCRIPT_NAME="MerlinAU"
 
 ##-------------------------------------##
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
@ExtremeFiretop,

Well, it turned out that porting the code I had written before for email notifications was easier than I thought so the integration & testing were completed faster than my initial estimate. So far it's working well on side of the fence.

If you have time, please try to set up email notifications via AMTM first, and then enable the option in the main menu so you can test the feature as you run your own test cycles.

BTW, I put the new option at the top main menu.  I'm not quite sure whether it should be on the new "Advance Options" menu. I really have no personal preference at this point so either way would be OK with me. The option is not visible at all unless you have AMTM optional feature enabled as well.

ExtremeFiretop:
> @ExtremeFiretop,
> 
> Well, it turned out that porting the code I had written before for email notifications was easier than I thought so the integration & testing were completed faster than my initial estimate. So far it's working well on my side of the fence.
> 
> If you have time, please try to set up email notifications via AMTM first, and then enable the option in the main menu so you can test the feature as you run your own test cycles.
> 
> BTW, I put the new option at the top main menu. I'm not quite sure whether it should be on the new "Advanced Options" menu. I really have no personal preference at this point so either way would be OK with me. The option is not visible at all unless you have AMTM optional feature enabled as well.

I'm on it! Test is my middle man, after broken that is. 😆😉

ExtremeFiretop:
I think I would recommend leaving this setting out in the open on the main menu if configured in AMTM.

We have more advanced options coming for the advanced menu anyways (like the option to enable beta to production firmware upgrades)

ExtremeFiretop:
@Martinski4GitHub 

It's perfect. I can't find any bugs. 
Coming from yours Truly, Extreme Test Firetop

ExtremeFiretop:
I think now that you implemented this, I'll take a different angle to advise for firmwares that need approval to flash due to the new changelog verification I just implemented.

Martinski4GitHub:
> @Martinski4GitHub
> 
> It's perfect. I can't find any bugs. Coming from yours Truly, Extreme Test Firetop
> 
Excellent!! That's what I like to hear :>) Home run on 1st try.

Martinski4GitHub:
> I think I would recommend leaving this setting out in the open on the main menu if configured in AMTM.
> 
> We have more advanced options coming for the advanced menu anyways (like the option to enable beta to production firmware upgrades)
>
That sounds good to me.

PR #91, opened 2024-01-26
Update MerlinAU.sh

Martinski4GitHub:
Added code to explicitly stop all Entware services, if Entware is found installed, before starting to flash the F/W image.


ExtremeFiretop:
My man! 

This is great, love solution, has it actually been tested through a flash yet or should I put it through it's paces?

Martinski4GitHub:
> My man!
> 
> This is great, love solution, has it actually been tested through a flash yet or should I put it through it's paces?
> I have not gone through the entire flash process yet, only to the point before the flash starts, but I can see the Entware services being stopped.

ExtremeFiretop:
Tested and confirmed functional.

PR #97, opened 2024-01-29
Update MerlinAU.sh

Martinski4GitHub:
Fixed a problem with getting the correct F/W version number to compare with an available update. The problem was related to the "beta" checking code. Also added "alpha" string checking just to be thorough and avoid issues with users using "alpha" versions wanting to straight to the production release.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -227,7 +227,7 @@ FW_UpdateCheckScript="/usr/sbin/webs_update.sh"
 _GetDefaultUSBMountPoint_()
 {
    local mounPointPath  retCode=0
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
BTW, I had to fix the problem in my local code because to was completely messing up my own testing.
I see that you fixed it in your code, but I also wanted to add the "alpha" checking.

ExtremeFiretop (approved):

ExtremeFiretop:
Approved. All looks good to me. Nice changes, nice job fixing it with regex, tighter than mine so we'll go with yours.

I'll still submit mine but yours wins for the conflict of the F/W

PR #104, opened 2024-01-29
Update MerlinAU.sh

Martinski4GitHub:
Improved the code that removes the "Alpha/Beta" suffixes from the version string.
I was testing the "Beta-to-Production" option and by chance set the Beta version like this:

"3004.386.12.4.Beta1"

Which gave this string:  "3004.386.12.4."

So I added possible separators for the "Beta" suffix. Now the resulting value is as expected:

"3004.386.12.4Beta1"   ==>  "3004.386.12.4"
"3004.386.12.4.Beta1"  ==>  "3004.386.12.4"
"3004.386.12.4_Beta1"  ==>  "3004.386.12.4"
"3004.386.12.4-Beta1"  ==>  "3004.386.12.4"

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Jan-28
+# Last Modified: 2024-Jan-29
 ###################################################################
 set -u
 
@@ -388,7 +388,7 @@ _FWVersionStrToNum_()
     then
         nonProductionVersionWeight=-100
         # Remove 'alpha/beta' and any following numbers #
-        verStr="$(echo "$verStr" | sed 's/[Aa]lpha[0-9]*// ; s/[Bb]eta[0-9]*//')"
+        verStr="$(echo "$verStr" | sed 's/[._-]\?[Aa]lpha[0-9]*// ; s/[._-]\?[Bb]eta[0-9]*//')"
     fi
 
     numFields="$(echo "$verStr" | awk -F '.' '{print NF}')"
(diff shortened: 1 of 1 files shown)


ExtremeFiretop (approved):

ExtremeFiretop:
Testing now

ExtremeFiretop:
Morning @Martinski4GitHub 

Works as intended, however using the following format: 3004.386.12.4Beta1 (outside the standard)  in the debug mode breaks the changelog verification:

Not sure if that's something we need to address next.

ExtremeFiretop:
> Morning @Martinski4GitHub
> 
> Works as intended, however using the following format: 3004.386.12.4.Beta1 (outside the standard) in the debug mode breaks the changelog verification:
> 
> Not sure if that's something we need to address next.

Seems to me that the reason is the matching becomes broken, it currently searchs the file for the startpoint using the standard format (3004.388.6) and then the end by searching for the next point following that format, and then parses that chunk between the 2 points.

But 3004.386.12.4.Beta1 (or 3004.386.12.4) for that matter, does not match the format we are looking for in the changelogs and due to that, checks the entire change log.

ExtremeFiretop:
I see 2 routes to the above problem, one is we simply disable changelog verification if those 2 points aren't found, or we try to adapt the matching so it can find formats such as: "3004.386.12.4"

I'm pretty sure (like 80% sure) my current solution would not match that format correctly now even if was in the file.

ExtremeFiretop:
> aren't found, or we try to adapt the matching so it can find formats such as: "3004.386.12.4"
> 
> I'm pretty sure (like 80% sure) my current solution would not match that format correctly now even if was in the file.

Agreed, I didn't sleep last night at all really, so just starting to look into this now LOL.
But I agree, I actually like both routes described.

1 is adjust the logic for the formats, and if after adjusting the logic for multiple formats still doesn't find it, then we just flat out disable it so people aren't confused.

PR #121, opened 2024-02-13
Update MerlinAU.sh

Martinski4GitHub:
Code improvement for parsing & validating the CRON schedule parameters to be able to accept name strings (e.g. Sat, Sun) for the day(s) of the week.


ExtremeFiretop:
This is good, I actually noticed we couldn't use SUN in the cron for example it was considered invalid even though technically the cron calculator could provide that as a valid cron.

I didn't look into it more but I had only noticed it while doing the translation to a readable format!

You caught it and fixed it before I got a chance to look into it next hehe

ExtremeFiretop (approved):

PR #125, opened 2024-02-17
Update MerlinAU.sh Backupmon Versioning Update

ExtremeFiretop:
1. **Remove the migrate settings function.** 
It's served it's purpose. It's been around since the pre-release version of 0.2.54 and we are now on 1.0.3 and going to 1.0.4, that means it's been in the script for at least **5** version cycles. **(0.2.54 --> 1.0.0 --> 1.0.1 --> 1.0.2 --> 1.0.3)**

2. **Upped BACKUPMON minimum version**
Increased the minimum version to the current of **1.5.3** using the new version standard going forwards. Considering he said in the forums him and thelonelycoder are also using this new standard going forwards, I figured it's best we up the minimum, should be one of the last times we need to do it.

3. **Added a method** for the script to still process the older versions such as **1.46** and **1.44** standards for comparison and treat them as **lower**.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -834,6 +834,87 @@ Update_Custom_Settings()
 }
 
 ##------------------------------------------##
+## Modified by ExtremeFiretop [2024-Jan-30] ##
+##------------------------------------------##
+_migrate_settings_move() {
+    local USBMountPoint="$(Get_Custom_Setting FW_New_Update_LOG_Directory_Path)"
+    local old_settings_dir="${ADDONS_PATH}/$ScriptFNameTag"
+    local new_settings_dir="${ADDONS_PATH}/$ScriptDirNameD"
+    local old_bin_dir="/home/root/$ScriptFNameTag"
+    local new_bin_dir="/home/root/$ScriptDirNameD"
+    local old_log_dir="${USBMountPoint}/$ScriptFNameTag"
+    local new_log_dir="${USBMountPoint}/$ScriptDirNameD"
+
+    # Check if the old SETTINGS directory exists #
+    if [ -d "$old_settings_dir" ]; then
+        # Check if the new SETTINGS directory already exists
+        if [ -d "$new_settings_dir" ]; then
+            # Remove the old SETTINGS directory since the new one exists
+            rm -rf "$old_settings_dir"
+            if [ $? -eq 0 ]; then
+                echo "The new SETTINGS directory already exists. Removed the old SETTINGS directory."
+            else
+                echo "Error occurred while removing the old SETTINGS directory."
+            fi
+        else
+            # Move the old SETTINGS directory to the new location
+            mv "$old_settings_dir" "$new_settings_dir"
+            if [ $? -eq 0 ]; then
+                echo "SETTINGS directory successfully migrated to the new location."
+            else
+                echo "Error occurred during migration of the SETTINGS directory."
+            fi
+        fi
+    fi
+
+    # Check if the old BIN directory exists #
+    if [ -d "$old_bin_dir" ]; then
+        # Check if the new BIN directory already exists
+        if [ -d "$new_bin_dir" ]; then
+            # Remove the old BIN directory since the new one exists
+            rm -rf "$old_bin_dir"
+            if [ $? -eq 0 ]; then
+                echo "The new BIN directory already exists. Removed the old BIN directory."
+            else
+                echo "Error occurred while removing the old BIN directory."
+            fi
+        else
+            # Move the old BIN directory to the new location
+            mv "$old_bin_dir" "$new_bin_dir"
+            if [ $? -eq 0 ]; then
+                echo "BIN directory successfully migrated to the new location."
+            else
+                echo "Error occurred during migration of the BIN directory."
+            fi
+        fi
+    fi
+
+    # Check if the old LOG directory exists #
+    if [ -d "$old_log_dir" ]; then
+        # Check if the new LOG directory already exists
+        if [ -d "$new_log_dir" ]; then
+            # Remove the old LOG directory since the new one exists
+            rm -rf "$old_log_dir"
+            if [ $? -eq 0 ]; then
+                echo "The new LOG directory already exists. Removed the old LOG directory."
+            else
+                echo "Error occurred while removing the old LOG directory."
+            fi
+        else
+            # Move the old LOG directory to the new location
+            mv "$old_log_dir" "$new_log_dir"
+            if [ $? -eq 0 ]; then
+                echo "LOG directory successfully migrated to the new location."
+            else
+                echo "Error occurred during migration of the LOG directory."
+            fi
+        fi
+    fi
+}
+
+_migrate_settings_move
+
+##------------------------------------------##
 ## Modified by ExtremeFiretop [2024-Jan-24] ##
 ##------------------------------------------##
 _Set_FW_UpdateLOG_DirectoryPath_()
@@ -1557,7 +1638,7 @@ _GetCurrentFWInstalledLongVersion_()
 _GetCurrentFWInstalledShortVersion_()
 {
 ##FOR DEBUG ONLY##
-if true
+if false
 then
     local theVersionStr  extVersNum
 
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
@Martinski4GitHub 

I actually just realized the last time we modified the migrate settings function was Jan 30th in version: 1.0.1
So it's still been in the script in its current form for at least 3 cycles, and an additional 2 cycles in a simpler form.

Considering it's only 3 cycles I'm open to leaving it in if you have a preference.
Let me know, I can easily commit it back in before we merge :)

MerlinAU.sh
```diff
@@ -2670,18 +2589,28 @@ Please manually update to version $minimum_supported_version or higher to use th
...
 
-            # Compare current version with the required version
+            # Adjust version format from 1.46 to 1.4.6 if needed
+            DOT_COUNT=$(echo "$BM_VERSION" | tr -cd '.' | wc -c)
+            if [ "$DOT_COUNT" -eq 0 ]; then
+                # If there's no dot, it's a simple version like "1" (unlikely but let's handle it)
+                BM_VERSION="${BM_VERSION}.0.0"
+            elif [ "$DOT_COUNT" -eq 1 ]; then
+                # For versions like 1.46, insert a dot before the last two digits
+                BM_VERSION=$(echo "$BM_VERSION" | sed 's/\.\([0-9]\)\([0-9]\)/.\1.\2/')
```
Martinski4GitHub, line 2606: I don't understand why you're converting the "2-dot" version string like "**1.46**" to "**1.4.6**" because the much simpler & accurate method is to convert it to "**1.46.0**" instead, which is the equivalent value when using the new version scheme (i.e. "**major.minor**"  --->  "**major.minor.0**").

The following code should do the job for the "2-dot" version:
`BM_VERSION="${BM_VERSION}.0"`

My 2 cents.

Martinski4GitHub:
> @Martinski4GitHub
> 
> I actually just realized the last time we modified the migrate settings function was Jan 30th in version: 1.0.1 So it's still been in the script in its current form for at least 3 cycles, and an additional 2 cycles in a simpler form.
> 
> Considering it's only 3 cycles I'm open to leaving it in if you have a preference. Let me know, I can easily commit it back in before we merge :)
>
I'd say leave the code for one more release cycle. It doesn't hurt since it's a simple check, IMO.

MerlinAU.sh
Martinski4GitHub: > I don't understand why you're converting the "2-dot" version string like "**1.46**" to "**1.4.6**" because the much simpler & accurate method is to convert it to "**1.46.0**" instead, which is the equivalent value when using the new version scheme (i.e. "**major.minor**" ---> "**major.minor.0**").
> 
> The following code should do the job for the "2-dot" version: `BM_VERSION="${BM_VERSION}.0"`
> 
> My 2 cents.
>
Ah, I see why now. BackupMon version went from "1.46" to  "1.5.0"  which is very strange and not conforming to well-established version scheme standards. IOW, "1.46" really means "1.46.0" so the next version should have been "1.47.0" or "1.50.0" In any case, I understand the need to adjust even if it was made due to an error on Viktor's side.

ExtremeFiretop: Hey @Martinski4GitHub 

Sorry it's been a crazy day! I'm here now though :)

Yes, exactly. The standard changed from 1.46 to .1.5.3 and when converted through our _ScriptVersionStrToNum_ function it would come out looking like: 1046000 instead of 1.4.6 which came out looking like: 1004006

When compared to: 1.5.3 and it came out like: 1005003 so if I didn't add the new not, it would compare : 1046000 vs 1005003 

1046000 is bigger than:
1005003 

and so it would never think the new standard was bigger (1.5.3)
However if I do what I currently do, and format it the same, then it properly compares as: 

1004006 vs
1005003 

And so it properly considers 1.5.3 as larger than 1.46

I know you already clued into all of this, but I just want it in writing in case we ever need to come back to this PR for some reason :)

ExtremeFiretop:
> > @Martinski4GitHub
> > I actually just realized the last time we modified the migrate settings function was Jan 30th in version: 1.0.1 So it's still been in the script in its current form for at least 3 cycles, and an additional 2 cycles in a simpler form.
> > Considering it's only 3 cycles I'm open to leaving it in if you have a preference. Let me know, I can easily commit it back in before we merge :)
> 
> I'd say leave the code for one more release cycle. It doesn't hurt since it's a simple check, IMO.

Agreed, I had a crazy day but today while I was out shopping with the girlfriend, she was trying clothes on and all I was thinking about was if this is safe to remove yet lol.

My conclusion is maybe someone installed our re-release and took a month long vacation, we haven't been a full month since the last time we touched it in 1.0.1 and so it's just safer to leave it than to take it out.

ExtremeFiretop:
No harm in leaving, still a small chance of potential harm to remove it.
So lets keep it in :) BRB! Back at my desk finally to commit it back in.

ExtremeFiretop:
@Martinski4GitHub 

Done. Looks right to me :) Feel free to review and merge when ready!

Martinski4GitHub:
> 
> My conclusion is maybe someone installed our re-release and took a month long vacation, we haven't been a full month since the last time we touched it in 1.0.1 and so it's just safer to leave it than to take it out.
>
Yes, exactly. The latest code hasn't been there long enough to allow all users to update, especially since this is not the kind of script that people run frequently in an interactive session. So for those month-long stragglers :>), let's give them more time to catch up.

ExtremeFiretop:
@Martinski4GitHub 

Are we pretty much done/ready for the next major release or did you have any other last minute thoughts of things to add/change?

I figured will have to do a small "maintenance" release after the next firmware drop to up the minimum firmware version supported. 

We can always re-assess the removal of the migration function in a maintenance release in the future.

Martinski4GitHub:
> @Martinski4GitHub
> 
> Are we pretty much done/ready for the next major release or did you have any other last minute thoughts of things to add/change?
> 
Sorry, my brother stopped by & we're going out for dinner soon.

Just quickly, he also likes the new feature of having a **CC** option for email notifications; but he also proposed to add an option to set the email format type to either "HTML" or "Plain Text."

The reason is two-fold: a) It gives users an "out" if their specific email server appears to mangle the emails being sent in HTML format, and b) If a user chooses to add the email address from the cellular carrier (e.g. T-Mobile, Verizon) to get SMS notifications, the "Plain Text" format would be the preferred choice.

I like his reasoning so I'm thinking of adding a menu option for that. What do you think? 
There's no rush for it, so we could release what we have now, and add the new option for the next version.

BTW, I was planning to make some minor changes to the email code (15 minutes tops with testing). I can do it after we come back from dinner.

Talk to you later.

ExtremeFiretop:
> > @Martinski4GitHub
> > Are we pretty much done/ready for the next major release or did you have any other last minute thoughts of things to add/change?
> 
> Sorry, my brother stopped by & we're going out for dinner soon.
> 
> Just quickly, he also likes the new feature of having a **CC** option for email notifications; but he also proposed to add an option to set the email format type to either "HTML" or "Plain Text."
> 
> The reason is two-fold: a) It gives users an "out" if their specific email server appears to mangle the emails being sent in HTML format, and b) If a user chooses to add the email address from the cellular carrier (e.g. T-Mobile, Verizon) to get SMS notifications, the "Plain Text" format would be the preferred choice.
> 
> I like his reasoning so I'm thinking of adding a menu option for that. What do you think? There's no rush for it, so we could release what we have now, and add the new option for the next version.
> 
> BTW, I was planning to make some minor changes to the email code (15 minutes tops with testing). I can do it after we come back from dinner.
> 
> Talk to you later.

No worries buddy enjoy dinner!
My girlfriend is actually just making dinner for us now. 

(Yes it's 11pm and it's later than usual for us, but like I said it's been a crazy day!)

I actually like the idea too. But take your time, I mean like you said this current solution we have we can expect to work for most major mail servers in north America, but your point and his point still stands that there can be a specific provider we don't account for, etc. and we should give those specific exceptions a way out.

It's one of those services that it's hard to account for all the variables and variations out in the world.

Martinski4GitHub:
> 
> No worries buddy enjoy dinner! My girlfriend is actually just making dinner for us now.
> 
> (Yes it's 11pm and it's later than usual for us, but like I said it's been a crazy day!)
> 
We've been known to have late dinners ourselves; 9:00 pm is fairly common, and on the weekends it could be around 11:00 pm while watching a couple of movies.

>
> I actually like the idea too. But take your time, I mean like you said this current solution we have we can expect to work for most major mail servers in north America, but your point and his point still stands that there can be a specific provider we don't account for, etc. and we should give those specific exceptions a way out.
> 
> It's one of those services that it's hard to account for all the variables and variations out in the world.
>
Yes, exactly. Having both format options at the control of the user would help in those unknown situations.

PR #159, opened 2024-03-03
Bug fix in Password input function

Martinski4GitHub:
There was a bug in the password input code that was suppressing escape sequences which also ended up suppressing chars right after pressing the Escape key (`<ESC>`) followed by the Left Square Bracket (`[`). For example: "`<ESC>[1`"

IOW, if a user were to type the keypress pattern "`<ESC>[X`" where `X` indicates any printable char, the Left Square Bracket and the following char were also suppressed.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -2002,7 +2002,7 @@ _GetPasswordInput_()
           if [ "$timeDiff" -gt 0 ]
           then
               showPSWD="$((! showPSWD))"
-              lastTabTime=$currentTime  # Update last TAB press time #
+              lastTabTime="$currentTime"  # Update last TAB press time #
               _showPSWDPrompt_
           fi
           continue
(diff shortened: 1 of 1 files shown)


ExtremeFiretop (approved):

ExtremeFiretop:
Approved. Good find/catch!

Martinski4GitHub:
> Approved. Good find/catch!
>
Last Saturday evening, I decided to take another look at the code handling escape sequences. I was not really satisfied with how I was suppressing them and had a nagging feeling that something was not quite right. So I re-tested the function several times and everything worked OK, but then it occurred to me that I should try to "simulate" an escape sequence by pressing the `<ESC>` key followed by the Left Square Bracket, and that's how I found the bug.

I don't think this bug fix requires an urgent new release version. This keypress combination `<ESC>[ ` is likely a rare occurrence while inputting your password; but, OTOH, the probability is not zero. In any case, if a user finds the bug within the next few days, a release can be issued very soon after that.

PR #163, opened 2024-03-16
Improved Memory Management

ExtremeFiretop:
Any thoughts on this solution Martinski?

Improved Memory Management in response to: 
https://www.snbforums.com/threads/merlinau-v1-0-7-the-ultimate-firmware-auto-updater-now-available-in-amtm.88577/post-897839

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -1836,7 +1836,7 @@ check_memory_and_prompt_reboot() {
 
             sync; echo 2 > /proc/sys/vm/drop_caches
 
-        	# Check available memory again #
+            # Check available memory again #
             availableRAM_kb=$(_GetAvailableRAM_KB_)
             if [ "$availableRAM_kb" -lt "$required_space_kb" ]; then
 
@@ -1858,7 +1858,7 @@ check_memory_and_prompt_reboot() {
 
                 sync; echo 3 > /proc/sys/vm/drop_caches
 
-        		# Check available memory again #
+                # Check available memory again #
                 availableRAM_kb=$(_GetAvailableRAM_KB_)
                 if [ "$availableRAM_kb" -lt "$required_space_kb" ]; then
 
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
Improvements with this design:

1. Improved messaging, before if the drop caches didn't work, we didn't see how much memory was available after it tried, now we do.
2. Also attempts to clear the dentries and inodes as a next progressive step if the simple Page Cache clear didn't work.
3. Finally as a final attempt it will shutdown entware and diversion (both fairly insensive on memory) and do a clear of everything, page cache, dentries and inodes.
4. If none of that worked, it then continues like usual to ask for a reboot, if it's not interactive, it will restart the services it stopped.

Martinski4GitHub:
> Improvements with this design:
> 
>     1. Improved messaging, before if the drop caches didn't work, we didn't see how much memory was available after it tried, now we do.
> 
>     2. Also attempts to clear the dentries and inodes as a next progressive step if the simple Page Cache clear didn't work.
> 
>     3. Finally as a final attempt it will shutdown entware and diversion (both fairly intensive on memory) and do a clear of everything, page cache, dentries and inodes.
> 
>     4. If none of that worked, it then continues like usual to ask for a reboot, if it's not interactive, it will restart the services it stopped.
>
Excellent job! I like the additional strategy to try to free up more memory by progressively increasing the chances and shutting down Entware services & Diversion at that point. If that's not sufficient to get enough available RAM, it could be that the router has too many other processes that are using a lot of RAM, in which case I don't see anything else we can do programmatically.

ExtremeFiretop:
> > Improvements with this design:
> > ```
> > 1. Improved messaging, before if the drop caches didn't work, we didn't see how much memory was available after it tried, now we do.
> > 
> > 2. Also attempts to clear the dentries and inodes as a next progressive step if the simple Page Cache clear didn't work.
> > 
> > 3. Finally as a final attempt it will shutdown entware and diversion (both fairly intensive on memory) and do a clear of everything, page cache, dentries and inodes.
> > 
> > 4. If none of that worked, it then continues like usual to ask for a reboot, if it's not interactive, it will restart the services it stopped.
> > ```
> 
> If that's not sufficient to get enough available RAM, it could be that the router has too many other processes that are using a lot of RAM, in which case I don't see anything else we can do programmatically.

Agreed, this is pretty much my last/best effort at memory management at a universal level, otherwise it's likely the router has too little RAM for the amount of services running and I won't be nit-picking each service.

Thank you for the review! A second mind to bounce ideas off makes the world of difference. I hope your enjoying your Friday and have a great weekend!

Martinski4GitHub:
> > 
> > If that's not sufficient to get enough available RAM, it could be that the router has too many other processes that are using a lot of RAM, in which case I don't see anything else we can do programmatically.
> 
> Agreed, this is pretty much my last/best effort at memory management at a universal level, otherwise it's likely the router has too little RAM for the amount of services running and I won't be nit-picking each service.
> 
Correct. In such situations, it becomes the responsibility of the user to figure out what services to stop and perhaps do an assessment of why those extra processes are running and taking that much RAM, and whether they are really needed on a router that has only 512MB RAM like the RT-AX86S.

>
> Thank you for the review! A second mind to bounce ideas off makes the world of difference. 
>
Indeed. Code reviews performed by a "2nd pair of eyes" is a s/w dev. concept that has proven to be effective at reducing the overall number of new bugs, and at increasing knowledge sharing since it's also considered a very good teaching tool. In the company I work for, all Agile teams are encouraged to do code reviews on a regular basis, especially when a new person joins the team, but also when major architectural/design changes are being made.

>
> I hope your enjoying your Friday and have a great weekend!
>
Thanks. I hope you do as well. Take care, buddy.

PR #167, opened 2024-03-19
Update Memory Management

ExtremeFiretop:
Update Memory Management

PR #169, opened 2024-03-22
More Improvements for F/W Update Notification

Martinski4GitHub:
More code improvements when displaying the F/W Update Notification "date+time" and for calculating when the next F/W Update can begin to occur based on the number of postponed days, adjusting for the hour discrepancy when Daylight Saving Time occurs between the last F/W Notification "date+time" stamp and the current time.

Earlier this evening it dawned on me that a cron job scheduling time resolution is in minutes so the F/W Update Notification time should be in minutes as well, so I have adjusted accordingly the "date+time" stamp used and displayed to the user. This should provide users with more accurate info to see whether the next cron job run would be **before** or **after** the postponement threshold to actually do the F/W update.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Mar-20
+# Last Modified: 2024-Mar-21
 ###################################################################
 set -u
 
@@ -69,8 +69,10 @@ readonly tempEMailBodyMsg="/tmp/var/tmp/tempEMailBodyMsg.$$.TXT"
 readonly saveEMailInfoMsg="${SETTINGS_DIR}/savedEMailInfoMsg.SAVE.TXT"
 readonly theEMailDateTimeFormat="%Y-%b-%d %a %I:%M:%S %p %Z"
 
-cronCmd="$(which crontab) -l"
-[ "$cronCmd" = " -l" ] && cronCmd="$(which cru) l"
+if [ -z "$(which crontab)" ]
+then cronListCmd="cru l"
+else cronListCmd="crontab -l"
+fi
 
 ##----------------------------------------------##
 ## Added/Modified by Martinski W. [2024-Jan-06] ##
@@ -459,7 +461,7 @@ readonly CRON_MONTH_RegEx="$CRON_MONTH_NAMES([\/,-]$CRON_MONTH_NAMES)*|([*1-9]|1
 readonly FW_UpdateMinimumPostponementDays=0
 readonly FW_UpdateDefaultPostponementDays=15
 readonly FW_UpdateMaximumPostponementDays=60
-readonly FW_UpdateNotificationDateFormat="%Y-%m-%d_%H:00:00"
+readonly FW_UpdateNotificationDateFormat="%Y-%m-%d_%H:%M:00"
 
 readonly MODEL_ID="$(_GetRouterModelID_)"
 readonly PRODUCT_ID="$(_GetRouterProductID_)"
@@ -1833,7 +1835,7 @@ check_memory_and_prompt_reboot()
         then
             freeRAM_kb="$(get_free_ram)"
             Say "Required RAM: ${required_space_kb} KB - RAM Free: ${freeRAM_kb} KB - RAM Available: ${availableRAM_kb} KB"
-            
+
             # Attempt to clear dentries and inodes. #
             Say "Attempting to free up memory again more aggressively..."
             sync; echo 2 > /proc/sys/vm/drop_caches
@@ -1845,8 +1847,8 @@ check_memory_and_prompt_reboot()
             then
                 freeRAM_kb="$(get_free_ram)"
                 Say "Required RAM: ${required_space_kb} KB - RAM Free: ${freeRAM_kb} KB - RAM Available: ${availableRAM_kb} KB"
-                
-                # Attempt to clear clears pagecache, dentries, and inodes after shutting down services		
+
+                # Attempt to clear clears pagecache, dentries, and inodes after shutting down services
                 Say "Attempting to free up memory once more even more aggressively..."
 
                 # Stop Entware services before F/W flash #
@@ -2419,7 +2421,7 @@ _AddCronJobEntry_()
 
    cru a "$CRON_JOB_TAG" "$newSchedule $CRON_JOB_RUN"
    sleep 1
-   if $cronCmd | grep -qE "$CRON_JOB_RUN #${CRON_JOB_TAG}#$"
+   if $cronListCmd | grep -qE "$CRON_JOB_RUN #${CRON_JOB_TAG}#$"
    then
        retCode=0
        "$newSetting" && \
@@ -2434,10 +2436,10 @@ _AddCronJobEntry_()
 _DelCronJobEntry_()
 {
    local retCode
-   if $cronCmd | grep -qE "$CRON_JOB_RUN #${CRON_JOB_TAG}#$"
+   if $cronListCmd | grep -qE "$CRON_JOB_RUN #${CRON_JOB_TAG}#$"
    then
        cru d "$CRON_JOB_TAG" ; sleep 1
-       if $cronCmd | grep -qE "$CRON_JOB_RUN #${CRON_JOB_TAG}#$"
+       if $cronListCmd | grep -qE "$CRON_JOB_RUN #${CRON_JOB_TAG}#$"
        then
            retCode=1
            printf "${REDct}**ERROR**${NOct}: Failed to remove cron job [${GRNct}${CRON_JOB_TAG}${NOct}].\n"
@@ -2714,14 +2716,14 @@ _CheckNewUpdateFirmwareNotification_()
 }
 
 ##----------------------------------------##
-## Modified by Martinski W. [2024-Mar-20] ##
+## Modified by Martinski W. [2024-Mar-21] ##
 ##----------------------------------------##
 _CheckTimeToUpdateFirmware_()
 {
    if [ $# -eq 0 ] || [ -z "$1" ] || [ -z "$2" ]
    then echo "**ERROR** **NO_PARAMS**" ; return 1 ; fi
 
-   local notifyTimeSecs  postponeTimeSecs  currentTimeSecs
+   local notifyTimeSecs  postponeTimeSecs  currentTimeSecs  dstAdjustSecs  dstAdjustDays
    local fwNewUpdateNotificationDate  fwNewUpdateNotificationVers  fwNewUpdatePostponementDays
 
    _CheckNewUpdateFirmwareNotification_ "$1" "$2"
@@ -2739,16 +2741,32 @@ _CheckTimeToUpdateFirmware_()
    if [ "$fwNewUpdatePostponementDays" -eq 0 ]
    then return 0 ; fi
 
-   postponeTimeSecs="$((fwNewUpdatePostponementDays * 86400))"
    currentTimeSecs="$(date +%s)"
    notifyTimeStrn="$(echo "$fwNewUpdateNotificationDate" | sed 's/_/ /g')"
    notifyTimeSecs="$(date +%s -d "$notifyTimeStrn")"
+
+   #----------------------------------------------------------------------
+   # Adjust calculation of postponed days as elapsed time in seconds to
+   # account for the hour discrepancy when Daylight Saving Time happens.
+   # This way we can avoid a scenario where the F/W Update "date+time"
+   # threshold is set one hour *after* the scheduled cron job is set
+   # to run again and check if it's time to update the router.
+   #----------------------------------------------------------------------
+   if [ "$(date -d @$currentTimeSecs +'%Z')" = "$(date -d @$notifyTimeSecs +'%Z')" ]
+   then dstAdjustSecs=86400  #24-hour day is same as always#
+   else dstAdjustSecs=82800  #23-hour day only when DST happens#
+   fi
+   dstAdjustDays="$((fwNewUpdatePostponementDays - 1))"
+   if [ "$dstAdjustDays" -eq 0 ]
+   then postponeTimeSecs="$dstAdjustSecs"
+   else postponeTimeSecs="$(((dstAdjustDays * 86400) + dstAdjustSecs))"
+   fi
 
    if [ "$((currentTimeSecs - notifyTimeSecs))" -ge "$postponeTimeSecs" ]
    then return 0 ; fi
 
    upfwDateTimeSecs="$((notifyTimeSecs + postponeTimeSecs))"
-   upfwDateTimeStrn="$(date -d @$upfwDateTimeSecs +"%A, %Y-%b-%d %I:00 %p")"
+   upfwDateTimeStrn="$(date -d @$upfwDateTimeSecs +"%A, %Y-%b-%d %I:%M %p")"
 
    Say "The firmware update to ${GRNct}${2}${NOct} version is currently postponed for ${GRNct}${fwNewUpdatePostponementDays}${NOct} day(s)."
    Say "The firmware update is expected to occur on or after ${GRNct}${upfwDateTimeStrn}${NOct}, depending on when your cron job is scheduled to check again."
@@ -3899,7 +3917,7 @@ if [ "$FW_UpdateCheckState" -eq 1 ]
 then
     runfwUpdateCheck=true
     # Check if the CRON job already exists #
-    if ! $cronCmd | grep -qE "$CRON_JOB_RUN #${CRON_JOB_TAG}#$"
+    if ! $cronListCmd | grep -qE "$CRON_JOB_RUN #${CRON_JOB_TAG}#$"
     then
         logo
         # If CRON job does not exist, ask user for permission to add #


ExtremeFiretop:
Thanks @Martinski4GitHub !!

I'll review this shortly currently testing this issue: https://www.snbforums.com/threads/merlinau-v1-0-8-the-ultimate-firmware-auto-updater-now-available-in-amtm.88577/post-899098

Martinski4GitHub:
>
> I'll review this shortly currently testing this issue: https://www.snbforums.com/threads/merlinau-v1-0-8-the-ultimate-firmware-auto-updater-now-available-in-amtm.88577/post-899098
>
Running the cron job every 10 minutes will most likely run into an issue with the lock file because the lock is designed to keep "everyone out" for 10 minutes while the F/W Update is happening (or until it exits). This is by design because we don't want users to inadvertently start the script again in another terminal session, especially if/when they don't realize that it's already running automatically via the cron job. I don't see this as an issue. I highly doubt a user will run the cron every 10 minutes or less under normal circumstances. I'd say the minimum should be every 24 hours, maybe every 12 hours if really paranoid about getting updates; anything less is just unnecessary given that the built-in F/W check is run every 24 hours also (I believe it runs between 2am and 3am every day at some random time).

ExtremeFiretop:
> > I'll review this shortly currently testing this issue: https://www.snbforums.com/threads/merlinau-v1-0-8-the-ultimate-firmware-auto-updater-now-available-in-amtm.88577/post-899098
> 
> Running the cron job every 10 minutes will most likely run into an issue with the lock file because the lock is designed to keep "everyone out" for 10 minutes while the F/W Update is happening (or until it exits). This is by design because we don't want users to inadvertently start the script again in another terminal session, especially if/when they don't realize that it's already running automatically via the cron job. I don't see this as an issue. I highly doubt a user will run the cron every 10 minutes or less under normal circumstances. I'd say the minimum should be every 24 hours, maybe every 12 hours if really paranoid about getting updates; anything less is just unnecessary given that the built-in F/W check is run every 24 hours also (I believe it runs between 2am and 3am every day at some random times).

I'm having trouble running it as a cron job, at all.
Having dinner, will chat about it when I'm back

ExtremeFiretop:
Even though I've set the cron to every 5 or 10 minutes, and made sure there is no lock file, it only runs and generates a log on every hour?

ExtremeFiretop:
Not sure if it's every hour but it's for sure not what I set in the cron

ExtremeFiretop:
The problem seems to be with non-standard crons such as:

https://crontab.guru/#0/5_*_*_*_*

ExtremeFiretop:
It should be : */5 * * * * for every 5 minutes not 0/5 * * * *

But we accept 0/5 * * * * and it causes issues since it's non-standard. For us and our script that translates to every 5 minutes, but in reality, it shouldn't. This causes the confusion to users @Martinski4GitHub

ExtremeFiretop:
Hopefully I made that clear, I'm back from dinner btw so if I didn't please let me know where I can clarify :)

Martinski4GitHub:
> It should be : */5 * * * * for every 5 minutes not 0/5 * * * *
> 
> But we accept 0/5 * * * * and it causes issues since it's non-standard. For us and our script that translates to every 5 minutes, but in reality, it shouldn't. This causes the confusion to users @Martinski4GitHub
>
Yes, I see what you mean. I fully agree that "0/5 * * * *" is non-standard which means that it might work on some Linux systems but not on others. And the router's embedded system is likely even more restrictive about what the cron daemon accepts or how it may interpret such syntax. While the code is trying its best to catch syntax errors, this one is syntactically valid but open to different interpretations based on the version of cron daemon running.

I'd say this one falls into the column of "user error" and our advice should be: "Don't do that!" :>)

ExtremeFiretop:
> > It should be : */5 * * * * for every 5 minutes not 0/5 * * * *
> > But we accept 0/5 * * * * and it causes issues since it's non-standard. For us and our script that translates to every 5 minutes, but in reality, it shouldn't. This causes the confusion to users @Martinski4GitHub
> 
> Yes, I see what you mean. I fully agree that "0/5 * * * *" is non-standard which means that it might work on some Linux systems but not on others. And the router's embedded system is likely even more restrictive about what the cron daemon accepts or how it may interpret such syntax. While the code is trying its best to catch syntax errors, this one is syntactically valid but open to different interpretations based on the version of cron daemon running.
> 
> I'd say this one falls into the column of "user error" and our advice should be: "Don't do that!" :>)

I noticed the issue when I replicated his cron schedule of 0/10 * * * * 

But been trying to find a way to solve it logically without redoing the whole translation stuff... Lol.

I kinda agree the easy solution is tell the user "don't do that" I was thinking of maybe trying to find a watch to watch non-standadd crons and give a warning or something or just reject them.

Martinski4GitHub:
> 
> I noticed the issue when I replicated his cron schedule of 0/10 * * * *
> 
> But been trying to find a way to solve it logically without redoing the whole translation stuff... Lol.
> 
> I kinda agree the easy solution is tell the user "don't do that" I was thinking of maybe trying to find a way to catch non-standadd crons and give a warning or something or just reject them.
>
That's the kind of "rabbit hole" I didn't want to go in when I wrote the function to validate the cron schedule entry. As I noted then, there are probably hundreds of syntactically valid entries that can result in a schedule that will never work. 
Here are obvious examples:
0 0 31 Feb *
0 0 31 Apr *

But there are others that are not so obvious. My take is we can make a best-effort attempt to catch syntax errors, after that the user is still responsible for learning and using standard syntax & semantically correct cron schedules.

My 2 cents.

Martinski4GitHub:
> 
> I noticed the issue when I replicated his cron schedule of 0/10 * * * *
> 
I have the feeling that this valid syntax may actually work as intended:    `0-59/10  *  *  *  *`

ExtremeFiretop:
Sorry let me get this review out of the way, I just wanted to take some time to look into this report

MerlinAU.sh
```diff
@@ -69,8 +69,10 @@ readonly tempEMailBodyMsg="/tmp/var/tmp/tempEMailBodyMsg.$$.TXT"
 readonly saveEMailInfoMsg="${SETTINGS_DIR}/savedEMailInfoMsg.SAVE.TXT"
 readonly theEMailDateTimeFormat="%Y-%b-%d %a %I:%M:%S %p %Z"
 
-cronCmd="$(which crontab) -l"
-[ "$cronCmd" = " -l" ] && cronCmd="$(which cru) l"
+if [ -z "$(which crontab)" ]
```
ExtremeFiretop, line 72: Approved.

MerlinAU.sh
```diff
@@ -459,7 +461,7 @@ readonly CRON_MONTH_RegEx="$CRON_MONTH_NAMES([\/,-]$CRON_MONTH_NAMES)*|([*1-9]|1
 readonly FW_UpdateMinimumPostponementDays=0
 readonly FW_UpdateDefaultPostponementDays=15
 readonly FW_UpdateMaximumPostponementDays=60
-readonly FW_UpdateNotificationDateFormat="%Y-%m-%d_%H:00:00"
+readonly FW_UpdateNotificationDateFormat="%Y-%m-%d_%H:%M:00"
```
ExtremeFiretop, line 464: Approved.

Martinski4GitHub:
> Sorry let me get this review out of the way, I just wanted to take some time to look into this report
> 
I understand. No worries.

MerlinAU.sh
```diff
@@ -2739,16 +2741,32 @@ _CheckTimeToUpdateFirmware_()
...
    notifyTimeSecs="$(date +%s -d "$notifyTimeStrn")"
 
+   #----------------------------------------------------------------------
+   # Adjust calculation of postponed days as elapsed time in seconds to
+   # account for the hour discrepancy when Daylight Saving Time happens.
+   # This way we can avoid a scenario where the F/W Update "date+time"
+   # threshold is set one hour *after* the scheduled cron job is set
+   # to run again and check if it's time to update the router.
+   #----------------------------------------------------------------------
+   if [ "$(date -d @$currentTimeSecs +'%Z')" = "$(date -d @$notifyTimeSecs +'%Z')" ]
```
ExtremeFiretop, line 2755: Approved.

ExtremeFiretop:
@Martinski4GitHub 

Nice work on the DTS adjustment notification time based on whether Daylight Saving Time (DST) is in effect :)

ExtremeFiretop:
> > I noticed the issue when I replicated his cron schedule of 0/10 * * * *
> 
> I have the feeling that this valid syntax may actually work as intended: `0-59/10 * * * *`

Bingo, correct this does work as expected.

Martinski4GitHub:
>
> > I have the feeling that this valid syntax may actually work as intended: `0-59/10 * * * *`
> 
> Bingo, correct this does work as expected.
>
I suspected it would because that's a standard syntax for "step intervals" given a "start" and "end" range. But the clearer way is still "`*/10`" for every 10 minutes.

ExtremeFiretop:
> > > I have the feeling that this valid syntax may actually work as intended: `0-59/10 * * * *`
> > 
> > 
> > Bingo, correct this does work as expected.
> 
> I suspected it would because that's a standard syntax for "step intervals" given a "start" and "end" range. But the clearer way is still "`*/10`" for every 10 minutes.

I responded to his report on SNB now, merged this in, that should hopefully address his concerns moving forwards.

Martinski4GitHub:
> 
> Nice work on the DTS adjustment notification time based on whether Daylight Saving Time (DST) is in effect :)
>
Hopefully, this will help users every time we go through the DST hour change and avoid reports asking about why the cron job doesn't work in those specific situations.

ExtremeFiretop:
BTW any ideas what the heck happened here?:

https://www.snbforums.com/threads/merlinau-v1-0-8-the-ultimate-firmware-auto-updater-now-available-in-amtm.88577/post-898975

The users report to me on SNB PM was: "I decided to uninstall MerlinAU and try again. It worked the 2nd time. But I had another quirk. I had disabled MerlinAU while I was struggling with the date. When I went to re-enable it, I get the error /usr/bin/crontab -l: not found
Failed to add the cron job

But if I looked at the cron jobs, the MerlinAU entry was there.

When I uninstalled MerlinAU, and looked at the cron jobs again, the MerlinAU entry was still there.

After I re-installed it, I could change the time, and the cron job correctly reflected the new time."

Martinski4GitHub:
> 
> I responded to his report on SNB now, merged this in, that should hopefully address his concerns moving forwards.
>
Sounds good.
WRT to the cron schedule, perhaps in your FAQ section, there could be a warning about using standard syntax for cron schedules. I know that just saying "standard syntax" might not mean much to most users who are not proficient in Linux, but perhaps you can add an example like the one you had trouble with and what the standard syntax should be instead. Just a thought.

ExtremeFiretop:
> > I responded to his report on SNB now, merged this in, that should hopefully address his concerns moving forwards.
> 
> Sounds good. WRT to the cron schedule, perhaps in your FAQ section, there could be a warning about using standard syntax for cron schedules. I know that just saying "standard syntax" might not mean much to most users who are not proficient in Linux, but perhaps you can add an example like the one you had trouble with and what the standard syntax should be instead. Just a thought.

Done deal. Sold. Great idea, working on this now...

ExtremeFiretop:
> BTW any ideas what the heck happened here?:
> 
> https://www.snbforums.com/threads/merlinau-v1-0-8-the-ultimate-firmware-auto-updater-now-available-in-amtm.88577/post-898975
> 
> The users report to me on SNB PM was: "I decided to uninstall MerlinAU and try again. It worked the 2nd time. But I had another quirk. I had disabled MerlinAU while I was struggling with the date. When I went to re-enable it, I get the error /usr/bin/crontab -l: not found Failed to add the cron job
> 
> But if I looked at the cron jobs, the MerlinAU entry was there.
> 
> When I uninstalled MerlinAU, and looked at the cron jobs again, the MerlinAU entry was still there.
> 
> After I re-installed it, I could change the time, and the cron job correctly reflected the new time."

Is this why you changed the first few lines in this PR?

Martinski4GitHub:
> > BTW any ideas what the heck happened here?:
> > https://www.snbforums.com/threads/merlinau-v1-0-8-the-ultimate-firmware-auto-updater-now-available-in-amtm.88577/post-898975
> > The users report to me on SNB PM was: "I decided to uninstall MerlinAU and try again. It worked the 2nd time. But I had another quirk. I had disabled MerlinAU while I was struggling with the date. When I went to re-enable it, I get the error /usr/bin/crontab -l: not found Failed to add the cron job
> > But if I looked at the cron jobs, the MerlinAU entry was there.
> > When I uninstalled MerlinAU, and looked at the cron jobs again, the MerlinAU entry was still there.
> > After I re-installed it, I could change the time, and the cron job correctly reflected the new time."
> 
> Is this why you changed the first few lines in this PR?
>
Partially yes, but not specifically to address the problem. It's doing essentially the same thing as before, but in a clearer way (IMO), and the variable name is now more precise: **cronListCmd** is more specific for what was intended as opposed to **cronCmd**.

However, I still don't understand how this error is occurring: "`/usr/bin/crontab -l: not found`"
If the command "`which crontab`" returns an empty string the code would have used the `cru` cmd instead. But if it returns the command path (as it obviously did, based on his report), then the command actually exists and it should not fail. It is said to have failed with "**not found**" but that does not make sense for a built-in command when the path was found.

At this point, I have no explanation for why the error occurred.

ExtremeFiretop:
> > > BTW any ideas what the heck happened here?:
> > > https://www.snbforums.com/threads/merlinau-v1-0-8-the-ultimate-firmware-auto-updater-now-available-in-amtm.88577/post-898975
> > > The users report to me on SNB PM was: "I decided to uninstall MerlinAU and try again. It worked the 2nd time. But I had another quirk. I had disabled MerlinAU while I was struggling with the date. When I went to re-enable it, I get the error /usr/bin/crontab -l: not found Failed to add the cron job
> > > But if I looked at the cron jobs, the MerlinAU entry was there.
> > > When I uninstalled MerlinAU, and looked at the cron jobs again, the MerlinAU entry was still there.
> > > After I re-installed it, I could change the time, and the cron job correctly reflected the new time."
> > 
> > 
> > Is this why you changed the first few lines in this PR?
> 
> Partially yes, but not specifically to address the problem. It's doing essentially the same thing as before, but in a clearer way (IMO), and the variable name is now more precise: **cronListCmd** is more specific for what was intended as opposed to **cronCmd**.
> 
> However, I still don't understand how this error is occurring: "`/usr/bin/crontab -l: not found`" If the command "`which crontab`" returns an empty string the code would have used the `cru` cmd instead. But if it returns the command path (as it obviously did, based on his report), then the command actually exists and it should not fail. It is said to have failed with "**not found**" but that does not make sense for a built-in command when the path was found.
> 
> At this point, I have no explanation for why the error occurred.

Yeah I noticed you touched it up but I just didn't have an answer myself as to why he got that error. I made him re-test the which command and it gave returned result for both (same as mine though) and I don't have this issue.

So I figured maybe it was a fluke. Then seeing you touch up the code and make it "clearer" it made me think to bring this up.

ExtremeFiretop:
@Martinski4GitHub 

Btw tomorrow I'll probably test my PR through start to finish.

My plan of attack is this:
1. Downgrade my node (so I can clearly validate the update was successful)
2. Configure MerlinAU again with a 15 min cron schedule.
3. Next, I'll do a bit of math, I'll identify how much RAM I have free at that moment, and how big the extracted extracted zip is, and add those together to get a total expected RAM usage when the script runs.
4. Next, I'll fill the rest of "unused" RAM by filling the rest of the space with junk in the HOME directory.
5. Once I have junk in the home directory that fills the expected unused space. I'll run the script. In my math I'll leave about 40MB free.
6. I'll report the results and the logs to confirm it was successful even though the router will be in a low memory state around 40MB free.

Martinski4GitHub:
> 
> ... I made him re-test the which command and it gave returned result for both (same as mine though) and I don't have this issue.
> 
That's the puzzling part. If the command is found and exists in the system (based on the result of the `which` cmd), why did it fail on that specific call in his case? If that's a fluke, it's extremely odd since `crontab` is such a basic command for the OS. That would be like having the `tar` command fail with "not found" on some calls.

Martinski4GitHub:
> @Martinski4GitHub
> 
> Btw tomorrow I'll probably test my PR through start to finish.
> 
> My plan of attack is this:
> 
>     1. Downgrade my node (so I can clearly validate the update was successful)
> 
>     2. Configure MerlinAU again with a 15 min cron schedule.
> 
>     3. Next, I'll do a bit of math, I'll identify how much RAM I have free at that moment, and how big the extracted zip is, and add those together to get a total expected RAM usage when the script runs.
> 
>     4. Next, I'll fill the rest of "unused" RAM by filling the rest of the space with junk in the HOME directory.
> 
>     5. Once I have junk in the home directory that fills the expected unused space. I'll run the script. In my math I'll leave about 40MB free.
> 
>     6. I'll report the results and the logs to confirm it was successful even though the router will be in a low memory state around 40MB free.
>
Sounds like a good plan. I'm still thinking that there could be some non-critical services that could be shut down and terminated when the function that checks for "Required RAM" cannot get enough "Available RAM " so it has to stop Entware services. I recall a post from 3 or 4 years ago (so it's a vague memory) when RMerlin recommended someone to stop a couple of built-in services during some troubleshooting, and he said that those were not critical services. I'll try to search for the post to see if I can find it.

ExtremeFiretop:
> > @Martinski4GitHub
> > Btw tomorrow I'll probably test my PR through start to finish.
> > My plan of attack is this:
> > ```
> > 1. Downgrade my node (so I can clearly validate the update was successful)
> > 
> > 2. Configure MerlinAU again with a 15 min cron schedule.
> > 
> > 3. Next, I'll do a bit of math, I'll identify how much RAM I have free at that moment, and how big the extracted zip is, and add those together to get a total expected RAM usage when the script runs.
> > 
> > 4. Next, I'll fill the rest of "unused" RAM by filling the rest of the space with junk in the HOME directory.
> > 
> > 5. Once I have junk in the home directory that fills the expected unused space. I'll run the script. In my math I'll leave about 40MB free.
> > 
> > 6. I'll report the results and the logs to confirm it was successful even though the router will be in a low memory state around 40MB free.
> > ```
> 
> Sounds like a good plan. I'm still thinking that there could be some non-critical services that could be shut down and terminated when the function that checks for "Required RAM" cannot get enough "Available RAM" so it has to stop Entware services. I recall a post from 3 or 4 years ago (so it's a vague memory) when RMerlin recommended someone to stop a couple of built-in services during some troubleshooting, and he said that those were not critical services. I'll try to search for the post to see if I can find it.

What's funny is I kinda think I know what thread your talking about, and I even looked myself but had little luck finding it.

It's probably far down in the history at this point. Felt like around the time I started visiting SNB forums.

If you find services you know you want to shut down feel free to suggestion them, the issue for me is I asked this user to provide his htop outputnl from ScMerlin and nothing seemed to be extremely high or out of place.

To all we know it's could be files in the home directory causing higher than expected usage and it has nothing to do with services.

ExtremeFiretop:
@Martinski4GitHub 

147MB (Zip Size) + 50% (Overhead) = 220MB (Expected usage)

To Identify Available Memory...
Used : theMemAvailable_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemAvailable: /{print $2}')"
And : echo "$theMemAvailable_KB"
Between each run...

Basic Math:
557MB (Available) - 220MB (Expected usage) = 336.5MB (To use up) 
MINUS (-) 40MB (To leave available) = 297MB (Total junk to fill)


TEST RESULTS:

1st Pass:
251 MB (Available After junk) = Inflate Error. Failed with SHA256 Failure.

Decreased the junk size by 8MB to 289MB

2nd Pass:
261MB (Available after junk)  = Inflate Error. Failed with SHA256 Failure.

Decreased the junk size by 10MB to 279MB

3rd Pass:
272MB (Available after junk)  = Success. 

However, I'm no longer in a *very* low memory state.
Test results are a bit null now in my opinion, it seems the RAM disk may have a partition size limit, and because my zip contains 2 firmware files, it pushes me over the edge when I try to extract them both + the junk...
Based on the logs, I was always able to get the non-rog (pure) version of the firmware extracted normally first.
So I changed my MerlinAU configuration to use the non-rog version and hoped that it would pass the SHA256 checksum.


RE-TEST RESULTS WITH PURE FILE:

550.45MB (Available) - 220.5MB (Expected usage) = 330MB (To use up) 
MINUS (-) 40MB (To leave available) = 290MB (Total junk to fill)

1st Pass:
254 MB (Available After junk) = Success.

However, I noticed in the logs available memory was still too high.

Seems that due to my setup with multiple firmwares in the zips, I'll never be able to properly recreate this situation with the script. The best I can do is manual re-creation.

1. Upload the firmware file manually to the router.
2. Upload the junk file manually to the router,
3. Confirm I'm in a low memory state.
4. Run the flash command the script uses.

I'll try this shortly...

ExtremeFiretop:
Actually, as I sit here having dinner.

I realized I could just lower the overhead requirements in the script. I'll give that a try after.

Martinski4GitHub:
> 
> What's funny is I kinda think I know what thread your talking about, and I even looked myself but had little luck finding it.
> 
> It's probably far down in the history at this point. Felt like around the time I started visiting SNB forums.
> 
> If you find services you know you want to shut down feel free to suggestion them, the issue for me is I asked this user to provide his htop output from ScMerlin and nothing seemed to be extremely high or out of place.
> 

>
> To all we know it's could be files in the home directory causing higher than expected usage and it has nothing to do with services.
>
Well, I didn't find the exact post from RMerlin about the non-critical built-in services, but I did find a couple of posts that pointed to the **Notification Center** (NT) services as the ones I recall as non-critical. One is from RMerlin in Jun-2020, the other one is from ColinTaylor posted this week.

https://www.snbforums.com/threads/question-jffs-sys-nc-nt_center-db.64697/#post-593677

https://www.snbforums.com/threads/asus-rt-ac86u-merlin-current-version-386-12_6-jffs2-argh-no-free-space-left-for-gc-can-you-help-me-troubleshoot-this-please.89353/#post-898810

Note that in both cases, the problem was that **nt_center** was creating a large file on the JFFS partition, which likely means that the service may have also allocated a big chunk of RAM for its own heap & memory stack. As I understand it, the Notification Center services are related to AiProtection.

So I created the following function to terminate all the NT services and to stop the **conn_diag** service:


> _ShutDownNonCriticalServices_ ()
>  {

     for procName in nt_center nt_monitor nt_actMail
     do
         procNum="$(ps w | grep -w "$procName" | grep -cv "grep -w")"
         printf "$procName: [$procNum]\n"
         [ "$procNum" -gt 0 ] && killall -9 "$procName" && sleep 1
     done
     procNum="$(ps w | grep -w "conn_diag" | grep -cv "grep -w")"
     printf "conn_diag: [$procNum]\n"
     [ "$procNum" -gt 0 ] && service stop_conn_diag && sleep 1
> }

I tested it on my router and got back only ~**3.0MB** of free RAM. It's not much, but then again I'm not using AiProtection and my router is already running with plenty of "Available RAM" so the kernel has no need to evict & reclaim file-backed memory pages, and no other process is making a request for RAM pages. In a "low memory" condition, it's possible that the increase of free RAM may be more significant; even if it's only ~10MB that might be just enough to pass the "Required RAM" check.

What do you think?

ExtremeFiretop:
> > What's funny is I kinda think I know what thread your talking about, and I even looked myself but had little luck finding it.
> > It's probably far down in the history at this point. Felt like around the time I started visiting SNB forums.
> > If you find services you know you want to shut down feel free to suggestion them, the issue for me is I asked this user to provide his htop output from ScMerlin and nothing seemed to be extremely high or out of place.
> 
> > To all we know it's could be files in the home directory causing higher than expected usage and it has nothing to do with services.
> 
> Well, I didn't find the exact post from RMerlin about the non-critical built-in services, but I did find a couple of posts that pointed to the **Notification Center** (NT) services as the ones I recall as non-critical. One is from RMerlin in Jun-2020, the other one is from ColinTaylor posted this week.
> 
> https://www.snbforums.com/threads/question-jffs-sys-nc-nt_center-db.64697/#post-593677
> 
> https://www.snbforums.com/threads/asus-rt-ac86u-merlin-current-version-386-12_6-jffs2-argh-no-free-space-left-for-gc-can-you-help-me-troubleshoot-this-please.89353/#post-898810
> 
> Note that in both cases, the problem was that **nt_center** was creating a large file on the JFFS partition, which likely means that the service may have also allocated a big chunk of RAM for its own heap & memory stack. As I understand it, the Notification Center services are related to AiProtection.
> 
> So I created the following function to terminate all the NT services and to stop the **conn_diag** service:
> 
> > _ShutDownNonCriticalServices_ ()
> > {
> 
> ```
>  for procName in nt_center nt_monitor nt_actMail
>  do
>      procNum="$(ps w | grep -w "$procName" | grep -cv "grep -w")"
>      printf "$procName: [$procNum]\n"
>      [ "$procNum" -gt 0 ] && killall -9 "$procName" && sleep 1
>  done
>  procNum="$(ps w | grep -w "conn_diag" | grep -cv "grep -w")"
>  printf "conn_diag: [$procNum]\n"
>  [ "$procNum" -gt 0 ] && service stop_conn_diag && sleep 1
> ```
> 
> > }
> 
> I tested it on my router and got back only ~**3.0MB** of free RAM. It's not much, but then again I'm not using AiProtection and my router is already running with plenty of "Available RAM" so the kernel has no need to evict & reclaim memory pages, and no other process is making a request for RAM pages. In a "low memory" condition, it's possible that the increase of free RAM may be more significant; even if it's only ~10MB that might be just enough to pass the "Required RAM" check when combined with the current 40MB "cushion" already in place.
> 
> What do you think?

Sorry for the delay I took a nap 😴 Haha.
I think it's a great idea, if it's non vital and it gives us some memory back, it's probably worth adding! :)

ExtremeFiretop:
I also don't use AI protection so I feel i'd be in a similar boat as you, but I can test it out and see if it frees anymore for me!

ExtremeFiretop:
Now that I think about it thought, my side gig has AI Protection enabled and is always being scanned by "external forces" lol.

Might be a better candidate to try this on.

ExtremeFiretop:
@Martinski4GitHub 

Even loweing my overhead in the script to 5% and increasing the junk in the home directory does not help me sadly...
Less of the secondary .w file gets unzipped in such a case, so I'm battleing the scripts math again in a different area now when it readjusts the new requirements.

If the junk is high, the secondary .w is small, and the adjustment reflects that, if the junk is low, the secondary .w almost gets fully extracted, so again the adjustment reflects that.

Examples below:

2024-03-23 06:50:22 Archive:  /tmp/mnt/USB2/MerlinAU.d/GT-AXE11000_firmware/GT-AXE11000_firmware.zip
2024-03-23 06:50:23   inflating: GT-AXE11000_3004_388.6_2_pureubi.w
2024-03-23 06:50:23   inflating: Changelog-386.txt
2024-03-23 06:50:23   inflating: Changelog-NG.txt
2024-03-23 06:50:23   inflating: sha256sum.sha256
2024-03-23 06:50:23   inflating: GT-AXE11000_3004_388.6_2_rog_pureubi.w
2024-03-23 06:50:23 unzip: write: No space left on device
2024-03-23 06:50:23 unzip: inflate error
2024-03-23 06:50:23 -----------------------------------------------------------
**2024-03-23 06:50:25 Total size of files: 88625172 bytes**

2024-03-23 06:35:19 Archive:  /tmp/mnt/USB2/MerlinAU.d/GT-AXE11000_firmware/GT-AXE11000_firmware.zip
2024-03-23 06:35:19   inflating: GT-AXE11000_3004_388.6_2_pureubi.w
2024-03-23 06:35:19   inflating: Changelog-386.txt
2024-03-23 06:35:19   inflating: Changelog-NG.txt
2024-03-23 06:35:19   inflating: sha256sum.sha256
2024-03-23 06:35:19   inflating: GT-AXE11000_3004_388.6_2_rog_pureubi.w
2024-03-23 06:35:19 unzip: write: No space left on device
2024-03-23 06:35:19 unzip: inflate error
2024-03-23 06:35:19 -----------------------------------------------------------
**2024-03-23 06:35:22 Total size of files: 107618324 bytes**

2024-03-23 00:10:28 Archive:  /tmp/mnt/USB2/MerlinAU.d/GT-AXE11000_firmware/GT-AXE11000_firmware.zip
2024-03-23 00:10:28   inflating: GT-AXE11000_3004_388.6_2_pureubi.w
2024-03-23 00:10:28   inflating: Changelog-386.txt
2024-03-23 00:10:28   inflating: Changelog-NG.txt
2024-03-23 00:10:28   inflating: sha256sum.sha256
2024-03-23 00:10:28   inflating: GT-AXE11000_3004_388.6_2_rog_pureubi.w
2024-03-23 00:10:28 unzip: write: No space left on device
2024-03-23 00:10:28 unzip: inflate error
2024-03-23 00:10:28 -----------------------------------------------------------
**2024-03-23 00:10:31 Total size of files: 153956372 bytes**

No matter what, I always end up with an "RAM Available" of about ~80128 KB
Now what I noticed is for my router with only 80MB available and no swap setup, the flashes seem to fail.

The script goes all the way to the end, says flashing, and then 4 minutes later (yes a full 4 minutes later) it reboots the same version. This has been consistent across all my tests.

The second I stop adding the junk, no problem, it says flashing and it comes back online 2 minutes later with the right version.
So even though I am working against the logic of the script, it seems I don't need to push it any lower than 80MB free on my model and the flashes fail.

Nothing in the system logs that I can see, no errors, no "crashes". It just reboots itself back to the old version.
Is this because the router froze? is this because my node is garbage? idk. It's a bad test because it it isn't apples for apples, my router isn't actually low on memory, I need to manufactory low memory, and my zip contains 2 firmware files so the math comes out differently, in his case he managed with 40MB and a swap file setup.

I have 80MB free and no swap setup and it doesn't seem to work. My next best guess is to setup a swap and see if somehow that allows it work, maybe the difference is the swap setup even though in our script we are trying to avoid swapping?

Otherwise maybe the difference is between the models, my model may need more memory to flash for some reason?
Would be best to have a more apples to apples comparison with a router that actually is low memory with 512MB of RAM.

At this point the best I can probably do is try to re-create the setup with a swap and see if that allows it to flash successfully.
In which case we could make it a provision that for this "adjustment" logic to come into play, that you must have a swap file.

Otherwise my last shot would be to do the manual suggestion I made before and try the steps manually. Maybe I could hard code the adjustment logic for now in the script as an alternative as well.

ExtremeFiretop:
Another idea would be to just try to manufacture real high memory usage.

Load up all the scripts. Every script. Turn up all the cron jobs. Do it all. Load that thing for real instead of trying to do it through the ram disk.

Not sure what my next step is. I think I'll take another nap and sleep on it. I didn't realize this would be so hard to recreate consistently.

ExtremeFiretop:
Ah, but we unmount the USB before the flash so it can't be related to the swap space anyways... Back to dreaming... 😴

ExtremeFiretop:
> > What's funny is I kinda think I know what thread your talking about, and I even looked myself but had little luck finding it.
> > It's probably far down in the history at this point. Felt like around the time I started visiting SNB forums.
> > If you find services you know you want to shut down feel free to suggestion them, the issue for me is I asked this user to provide his htop output from ScMerlin and nothing seemed to be extremely high or out of place.
> 
> > To all we know it's could be files in the home directory causing higher than expected usage and it has nothing to do with services.
> 
> Well, I didn't find the exact post from RMerlin about the non-critical built-in services, but I did find a couple of posts that pointed to the **Notification Center** (NT) services as the ones I recall as non-critical. One is from RMerlin in Jun-2020, the other one is from ColinTaylor posted this week.
> 
> https://www.snbforums.com/threads/question-jffs-sys-nc-nt_center-db.64697/#post-593677
> 
> https://www.snbforums.com/threads/asus-rt-ac86u-merlin-current-version-386-12_6-jffs2-argh-no-free-space-left-for-gc-can-you-help-me-troubleshoot-this-please.89353/#post-898810
> 
> Note that in both cases, the problem was that **nt_center** was creating a large file on the JFFS partition, which likely means that the service may have also allocated a big chunk of RAM for its own heap & memory stack. As I understand it, the Notification Center services are related to AiProtection.
> 
> So I created the following function to terminate all the NT services and to stop the **conn_diag** service:
> 
> > _ShutDownNonCriticalServices_ ()
> > {
> 
> ```
>  for procName in nt_center nt_monitor nt_actMail
>  do
>      procNum="$(ps w | grep -w "$procName" | grep -cv "grep -w")"
>      printf "$procName: [$procNum]\n"
>      [ "$procNum" -gt 0 ] && killall -9 "$procName" && sleep 1
>  done
>  procNum="$(ps w | grep -w "conn_diag" | grep -cv "grep -w")"
>  printf "conn_diag: [$procNum]\n"
>  [ "$procNum" -gt 0 ] && service stop_conn_diag && sleep 1
> ```
> 
> > }
> 
> I tested it on my router and got back only ~**3.0MB** of free RAM. It's not much, but then again I'm not using AiProtection and my router is already running with plenty of "Available RAM" so the kernel has no need to evict & reclaim memory pages, and no other process is making a request for RAM pages. In a "low memory" condition, it's possible that the increase of free RAM may be more significant; even if it's only ~10MB that might be just enough to pass the "Required RAM" check when combined with the current 40MB "cushion" already in place.
> 
> What do you think?

Tested this function on both my router and my side gig router:

 for procName in nt_center nt_monitor nt_actMail
 do
     procNum="$(ps w | grep -w "$procName" | grep -cv "grep -w")"
     printf "$procName: [$procNum]\n"
     [ "$procNum" -gt 0 ] && killall -9 "$procName" && sleep 1
 done
 procNum="$(ps w | grep -w "conn_diag" | grep -cv "grep -w")"
 printf "conn_diag: [$procNum]\n"
 [ "$procNum" -gt 0 ] && service stop_conn_diag && sleep 1

To Identify Available Memory...
Used : theMemAvailable_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemAvailable: /{print $2}')"
And : echo "$theMemAvailable_KB"
Between each run...

My Router. (Before Script):
547960KB
My Router. (After Script):
550420KB

**(Freed 2.4MB)**

Side Gig Router with AIProtection. (Before Script):
431300KB
Side Gig Router with AIProtection. (After Script):
435856KB

**Freed (4.5MB)**

ExtremeFiretop:
So now I leave this up to you, if you think shutting these down is risk free/worth it or not.
As I'd do with my management, sometimes it's about what you DON't say... ;)

**Manager:**
Did it give back RAM? 

**Me:**
"Absolutely..."

**Manager:**
"In both cases?" 

**Me:**
"Yep!"

**Left unsaid in my head:**
...Barely nothing, but no one asked that one!

ExtremeFiretop:
Does Swap Make a Difference? - No... As expected since we unmount before flash.
Maybe I could hard code the adjustment logic for now in the script as an alternative. - Trying now.
Manufacture real high memory usage. - Next...
Manual suggestion - Next...

ExtremeFiretop:
Commented out these lines and will try my tests again:

```
        # Check if the adjusted required space is less than the minimum cushion
        #if [ "$required_space_kb" -lt "$minimum_cushion_kb" ]; then
        #    # Add the difference to fulfill the minimum cushion
        #    cushion_diff=$((minimum_cushion_kb - required_space_kb))
        #    required_space_kb=$((required_space_kb + cushion_diff))
        #    Say "Added cushion of $cushion_diff KB to meet the minimum required RAM of 40MB."
        #fi
```

ExtremeFiretop:
151MB (Zip Size) + 5% (Overhead) = 226.5MB (Expected usage) 

To Identify Available Memory...
Used : theMemAvailable_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemAvailable: /{print $2}')"
And : echo "$theMemAvailable_KB"
Between each run...

Basic Math:
550.5MB (Available) - 226.5MB (Expected usage) = 324MB (To use up) 
MINUS (-) 35MB (To leave available) = 289MB (Total junk to fill)


TEST RESULTS:

1st Pass:
254 MB (Available After junk) = Script was unable to clear enough memory, script exited right before flash.

Increase junk by 43MB to 332MB
Hard coded overhead requirements to "-75%" to lower them to about 37MB required. Re-adjustment logic is disabled so I don't need to fight it either. This will test the script straight up against a requirement of 37MB and low memory.

2nd Pass:
217 MB (Available After junk) = Flash Started with 80MB. Disconnected me after 1 minute but I was almost instantly able to reconnect. Router very clearly ran out of memory. Even the system logs stopped populating after this line in the system logs:

`Mar 23 20:00:32 [MerlinAU.sh] 7957: unzip: write: No space left on device`

Then at about the 4 minute mark, the router fully rebooted. Came back online with the same version.
I'm not going to push my router any harder in that direction, instead I'll start to slowly back off until it works.

ExtremeFiretop:
Decrease junk by 20MB to 312MB. Rest of the script remains the same.

3rd Pass:
232 MB (Available After junk) = The script started flashing, but somehow yet it still only has "80MB" available. 
Even after decreasing the junk size. It's like the router/script doesn't have the resolution to see any lower than 80MB for me.

After 4 minutes the router rebooted to the same version still.

Decrease junk by 20MB to 292MB. Rest of the script remains the same.

4th Pass:
251 MB (Available After junk) = The script started flashing with "78MB" RAM available somehow... 
Even though I'm decreasing the junk twice, somehow available memory is now lower. Then after 2 minutes the router disconnected me. I was still able to reconnect nearly instantly, after 2 minutes the router rebooted, came back online with the same version.

Decrease junk by 20MB to 272MB. Rest of the script remains the same.

5th Pass:
272 MB (Available After junk) = The script started flashing with "85MB" RAM available... Finally. It's going in the right direction.
Router rebooted after 1 minute. Came back online with the same version.

Decrease junk by 20MB to 252MB. Rest of the script remains the same.

6th Pass:
292 MB (Available After junk) = The script started flashing with "105MB" RAM available.. Router rebooted after 1 minute.
Came back online with the same version still.

Decrease junk by 20MB to 232MB. Rest of the script remains the same.

7th Pass:
313 MB (Available After junk) =  The script started flashing with "126MB" RAM available.. Router rebooted after 1 minute.
Came back online with the same version still.

Decrease junk by 20MB to 212MB. Rest of the script remains the same.

8th Pass:
333 MB (Available After junk) = The script started flashing with "146MB" RAM available.. Router rebooted after 1 minute.
Came back online successfully updated. Logs below:

```
2024-03-23 21:15:01 MerlinAU v1.0.9
2024-03-23 21:15:01 Running the update task now... Checking for F/W updates...
2024-03-23 21:15:02 Required RAM: 37879 KB - RAM Free: 345820 KB - RAM Available: 332764 KB
2024-03-23 21:15:02 Backup script (BACKUPMON) is not installed. Skipping backup.
2024-03-23 21:15:03 Latest release version is 3004.388.6.2.
2024-03-23 21:15:03 Downloading https://sourceforge.net/projects/asuswrt-merlin/files/GT-AXE11000/Release/GT-AXE11000_3004_388.6_2.zip
2024-03-23 21:15:18 Required RAM: 37879 KB - RAM Free: 189504 KB - RAM Available: 314352 KB
2024-03-23 21:15:18 -----------------------------------------------------------
2024-03-23 21:15:18 Archive:  /tmp/mnt/USB2/MerlinAU.d/GT-AXE11000_firmware/GT-AXE11000_firmware.zip
2024-03-23 21:15:18   Length      Date    Time    Name
2024-03-23 21:15:18 ---------  ---------- -----   ----
2024-03-23 21:15:18  85065748  02-26-2024 23:01   GT-AXE11000_3004_388.6_2_pureubi.w
2024-03-23 21:15:18      9645  02-26-2024 21:20   README-merlin.txt
2024-03-23 21:15:18     94437  02-26-2024 22:21   Changelog-386.txt
2024-03-23 21:15:18     99381  02-26-2024 21:20   Changelog-NG.txt
2024-03-23 21:15:18       206  02-26-2024 23:10   sha256sum.sha256
2024-03-23 21:15:18  87687188  02-26-2024 23:10   GT-AXE11000_3004_388.6_2_rog_pureubi.w
2024-03-23 21:15:18  --------                     -------
2024-03-23 21:15:18 172956605                     6 files
2024-03-23 21:15:18 -----------------------------------------------------------
2024-03-23 21:15:25 Archive:  /tmp/mnt/USB2/MerlinAU.d/GT-AXE11000_firmware/GT-AXE11000_firmware.zip
2024-03-23 21:15:25   inflating: GT-AXE11000_3004_388.6_2_pureubi.w
2024-03-23 21:15:25   inflating: Changelog-386.txt
2024-03-23 21:15:25   inflating: Changelog-NG.txt
2024-03-23 21:15:25   inflating: sha256sum.sha256
2024-03-23 21:15:25   inflating: GT-AXE11000_3004_388.6_2_rog_pureubi.w
2024-03-23 21:15:25 -----------------------------------------------------------
2024-03-23 21:15:29 Total size of files: 172752936 bytes
2024-03-23 21:15:29 Adjusted required RAM by subtracting sizes of .w and .pkgtb files: 168704 KB. New required RAM: 37879 KB
2024-03-23 21:15:29 Required RAM: 37879 KB - RAM Free: 28064 KB - RAM Available: 146724 KB
2024-03-23 21:15:29 No high-risk phrases found in the change-log.
2024-03-23 21:15:30 Required RAM: 37879 KB - RAM Free: 27900 KB - RAM Available: 146576 KB
2024-03-23 21:15:30 Pure Build selected for flashing
2024-03-23 21:15:30 Required RAM: 37879 KB - RAM Free: 27520 KB - RAM Available: 146396 KB
2024-03-23 21:15:30 Router Web URL is: http://192.168.50.254
2024-03-23 21:15:35 Flashing GT-AXE11000_3004_388.6_2_pureubi.w... Please wait for reboot in about 4 minutes or less.

```

ExtremeFiretop:
What this tells me is for some reason I need much more RAM than expected to FLASH.
Almost double the firmware file size of 83MB. This could be an issue with my specific router with it's flash memory issues.

Based on the behavior, the fact that it started rebooting after 1 minute and not 4, and not "freezing". Tells me it likely *should* of worked as of the 5th Pass. Maybe my damaged router needs more for some reason, but really the take away here is that I needed at least 85MB of RAM available for the router/script to not "freeze" and I also noticed that our code:

`theMemAvailable_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemAvailable: /{print $2}')"`

Doesn't seem to have the resolution to see any lower than 78MB for my router, any lower and it randomly sticks at 80MB free. almost like a "default" value.

ExtremeFiretop:
This tells me if we ever want to refine the memory management, to add the adjustment logic in the memory management function, we would need to make it for routers with 512MB or lower specifically. Or do what Viktor suggested and do a reboot to per MerlinAU at the start of the pack. But we can't use this generically/globally.

ExtremeFiretop:
I found this for a list of services:

https://www.snbforums.com/threads/the-service-command-and-how-does-it-work.19763/#post-142785

ExtremeFiretop:
Or this:

https://www.snbforums.com/threads/i-would-like-to-kill-some-of-the-processes-permanently.47616/

ExtremeFiretop:
@Martinski4GitHub 

Added your suggested changes to my PR and removed the code that worked for that user for the memory adjustment.
https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/167/commits/0993ac545f0d451398da2813323a2755958e9532

Martinski4GitHub:
>
> Nothing in the system logs that I can see, no errors, no "crashes". It just reboots itself back to the old version. Is this because the router froze? is this because my node is garbage? idk. It's a bad test because it it isn't apples for apples, my router isn't actually low on memory, I need to manufactory low memory, and my zip contains 2 firmware files so the math comes out differently, in his case he managed with 40MB and a swap file setup.
>
> ...
>
> Otherwise maybe the difference is between the models, my model may need more memory to flash for some reason? Would be best to have a more apples to apples comparison with a router that actually is low memory with 512MB of RAM.
> 
Yeah, since your router has 1GB RAM and the ZIP file contains 2 F/W images, you have additional factors to deal with, plus your router model has a different SDK version. All that combined might be why this model requires more "Available RAM" to do the F/W update when compared with the models having only 512MB RAM. There are more unknowns when trying to recreate the same "low memory" scenario with your router.

Martinski4GitHub:
> So now I leave this up to you, if you think shutting these down is risk free/worth it or not. This was your idea so I'll let you lead it till the end. As I'd say with my management at times, sometimes it's about what you DON'T say... ;)
> 
> **Manager:** "Did it give back RAM?"
> 
> **Me:** "Absolutely..."
> 
> **Manager:** "In both cases?"
> 
> **Me:** "Yep!"
> 
> **Left unsaid in my head:** ...Barely nothing, but no one asked that question!
>
I've been running my router *without* the 4 services executing since I ran the code to shut them down, and so far I have seen no ill effect at all in behavior, performance, or anything else affecting wired or wireless clients. For all intents & purposes, those services do appear to be non-critical, with the caveat that I do not use AiProtection. But considering that we're going to be shutting down services before the F/W flash, I would think AiProtection is not critical at all, especially since that's used for protecting HTTP traffic. IMO, I'd try let's try it for the next release.

Martinski4GitHub:
> This tells me if we ever want to refine the memory management, to add the adjustment logic in the memory management function, we would need to make it for routers with 512MB or lower specifically. Or do what Viktor suggested and do a reboot to push MerlinAU at the start of the pack. But we can't use this generically/globally.
>
Based on your test results, I fully agree. Let's not mess around with the logic for routers with 1GB RAM. Make the adjustments only if a router has 512MB RAM or less.

ExtremeFiretop:
> 
> 2nd Pass: 217 MB (Available After junk) = Flash Started with 80MB RAM available. Disconnected me after 1 minute but I was almost instantly able to reconnect. Router very clearly ran out of memory. Even the system logs stopped populating after this line in the system logs:
> 
> `Mar 23 20:00:32 [MerlinAU.sh] 7957: unzip: write: No space left on device`
> 
> Then at about the 4 minute mark, the router fully rebooted. Came back online with the same version. I'm not going to push my router any harder in that direction, instead I'll start to slowly back off until it works.


> 
> No matter what, I always end up with an "RAM Available" of about ~80128 KB Now what I noticed is for my router with only 80MB available and no swap setup, the flashes seem to fail.
> 
> The script goes all the way to the end, says flashing, and then 4 minutes later (yes a full 4 minutes later) it reboots the same version. This has been consistent across all my tests.
> 
> The second I stop adding the junk, no problem, it says flashing and it comes back online 2 minutes later with the right version. So even though I am working against the logic of the script, it seems I don't need to push it any lower than 80MB free on my model and the flashes fail.
> 
> Nothing in the system logs that I can see, no errors, no "crashes". It just reboots itself back to the old version. Is this because the router froze? is this because my node is garbage? idk. It's a bad test because it it isn't apples for apples, my router isn't actually low on memory, I need to manufactory low memory, and my zip contains 2 firmware files so the math comes out differently, in his case he managed with 40MB and a swap file setup.
> 
> I have 80MB free and no swap setup and it doesn't seem to work. My next best guess is to setup a swap and see if somehow that allows it work, maybe the difference is the swap setup even though in our script we are trying to avoid swapping?



> 
> Decrease junk by 20MB to 312MB. Rest of the script remains the same.
> 
> 3rd Pass: 232 MB (Available After junk) = The script started flashing, but somehow yet it still only has "80MB" available. Even after decreasing the junk size. It's like the router/script doesn't have the resolution to see any lower than 80MB for me.
> 
> After 4 minutes the router rebooted to the same version still.
> 
> Decrease junk by 20MB to 292MB. Rest of the script remains the same.
> 
> 4th Pass: 251 MB (Available After junk) = The script started flashing with "78MB" RAM available somehow... Even though I'm decreasing the junk twice, somehow available memory is now lower. Then after 2 minutes the router disconnected me. I was still able to reconnect nearly instantly, after 2 minutes the router rebooted, came back online with the same version.
> 
> Decrease junk by 20MB to 272MB. Rest of the script remains the same.
> 
> 5th Pass: 272 MB (Available After junk) = The script started flashing with "85MB" RAM available... Finally. It's going in the right direction. Router rebooted after 1 minute. Came back online with the same version.

Agreed, the only thing I kinda want to figure out now is what this code: theMemAvailable_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemAvailable: /{print $2}')" seems to work correctly for the users 512MB router but can't see the resolution/detail for my 1GB router any lower than 80MB as found in my tests above.

Is this an issue with my specific router? Should I be looking into this more?
I noticed I said I'd let you run with your change and then just merged it, but I didn't mean to merge it, I just wanted to add the latest changes from dev into my PR and it merged them all and closed them all.

Clearly I still need to learn a thing about Github, luckily I reviewed the merge and everything seemed to come out as expected after solving the conflict. You can feel free to remove/add that part of the code in another PR if you wish. :) I only touched it up a little bit to include the other services.

Martinski4GitHub:
> 
> Agreed, the only thing I kinda want to figure out now is what this code: `theMemAvailable_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemAvailable: /{print $2}')"` seems to work correctly for the users 512MB router but can't see the resolution/detail for my 1GB router any lower than 80MB as found in my tests above.
> 
> Is this an issue with my specific router? Should I be looking into this more? 
>
I'm not sure I fully understand what's happening. The code is reading the value directly from the "**/proc/meminfo**" which is part of the kernel's special pseudo-filesystem that provides access to the most current information of the running OS. You said you found out that the watermarks/thresholds for "low memory" management were the same as mine, so the "high" threshold is still 30MB even for 1GB RAM. However, it's possible that ASUS may have implemented some other kind of threshold above those from the kernel where the minimum is 80MB at which point they do garbage collection and memory page eviction & reclamation to make sure it doesn't go below that custom watermark. This is just speculation. Remember that we're dealing with several "black boxes" around this F/W Update process so we don't know about any of the custom internal mechanisms ASUS has done around memory management.

I don't think you need to worry about that unless it causes problems in an actual "low memory" situation, not when you're trying to "forcefully fake" one by filling up the **tmpfs** filesystem.

>
> I noticed I said I'd let you run with your change and then just merged it, but I didn't mean to merge it, I just wanted to add the latest changes from dev into my PR and it merged them all and closed them all.
> 
> Clearly I still need to learn a thing about Github, luckily I reviewed the merge and everything seemed to come out as expected after solving the conflict. You can feel free to remove/add to that part of the code in another PR if you wish. :) I only touched it up a little bit to include the other services.
>
Yeah, no worries. I'll take a look when I update my fork and review the changes.

ExtremeFiretop:
> > Agreed, the only thing I kinda want to figure out now is what this code: `theMemAvailable_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemAvailable: /{print $2}')"` seems to work correctly for the users 512MB router but can't see the resolution/detail for my 1GB router any lower than 80MB as found in my tests above.
> > Is this an issue with my specific router? Should I be looking into this more?
> 
> I'm not sure I fully understand what's happening. The code is reading the value directly from the "**/proc/meminfo**" which is part of the kernel's special pseudo-filesystem that provides access to the most current information of the running OS. You said you found out that the watermarks/thresholds for "low memory" management were the same as mine, so the "high" threshold is still 30MB even for 1GB RAM. However, it's possible that ASUS may have implemented some other kind of threshold above those from the kernel where the minimum is 80MB at which point they do garbage collection and memory page eviction & reclamation to make sure it doesn't go below that custom watermark. This is just speculation. Remember that we're dealing with several "black boxes" around this F/W Update process so we don't know about any of the custom internal mechanisms ASUS has done around memory management.
> 
> I don't think you need to worry about that unless it causes problems in an actual "low memory" situation, not when you're trying to "forcefully fake" one by filling up the **tmpfs** filesystem.
> 
> > I noticed I said I'd let you run with your change and then just merged it, but I didn't mean to merge it, I just wanted to add the latest changes from dev into my PR and it merged them all and closed them all.
> > Clearly I still need to learn a thing about Github, luckily I reviewed the merge and everything seemed to come out as expected after solving the conflict. You can feel free to remove/add to that part of the code in another PR if you wish. :) I only touched it up a little bit to include the other services.
> 
> Yeah, no worries. I'll take a look when I update my fork and review the changes.

That's exactly why I'm confused, and thought maybe I should investigate more and see if I can replicate with my primary router.

Or course for my primary I would fill with Junk and not follow through the flash, but just see if it can go below 80MB. But because I know the kernel watermarks and saw this other users router get down to the 30MBs free and flash, I figured I should be able to get that low as well.

I think I was, but for some reason the router and by extension, our script, was misreporting this for some reason.

I don't think it's an actual issue, but it makes me wondering if it's something with my specific node or not.

It's weird because by the data I collect before I start the script and only load the garbage, you can clearly see it takes the appropriate sized bite of of my RAM, yet when I run the script it still reports 80MB free... when I know for a fact I took more RAM out of the question (usually by 20MB chunks at a time).

So the router was probably running in a state lower than 80MB but for some reason it was always reporting 80MB free. If I have some time I may retest this on my primary, but for now it doesn't seem to impact the function since it only starts to lose accuracy at around 80MB or lower for me which is never going to happen my node usually runs bare bones and has over 557MB free as we speak.

Anyways take your time, review when you can, I included your changes for the services, plus the new toggle and remove the migration function. If you have no other ideas or concerns I'll release v1.0.9 sometime tomorrow or Monday.

Goodnight buddy!

ExtremeFiretop:
And to wrap up the thought, this is the 2 reasons why I removed my suggestion for the memory management from my PR.

1. I wasn't able to get a successful flash without being in the 140MB free range which is more in line with our current 50% overhead requirements.
2. Seeing this potential issue with routers miss reporting below 80MB doesn't give me the confidence that all router could use that adjustment logic correctly with the 40MB cushion we originally intended.

If we ever re-implement such logic, it would again be with a 40MB cushion for safety, and only for the 512MB models. For now we would need a bigger sample size, I tried. Trust me I did. But I can't replicate correctly so if another user has an issue down the line we can try and grow the sample size that way.

Martinski4GitHub:
> 
> I think I was, but for some reason the router and by extension, our script, was misreporting this for some reason.
> ...
> So the router was probably running in a state lower than 80MB but for some reason it was always reporting 80MB free. 
>
If the router's kernel is indeed reporting the wrong information for "Available RAM" that would be a major bug since other processes usually rely on information from "**/proc/meminfo**" to evaluate various states of current memory.

Anyway, have a good night's rest buddy. Looks like you certainly had a full day today running those tests back & forth and collecting data points to figure out what was happening. I admire your persistence & determination. My hat's off to you, brother!!

ExtremeFiretop:
> > I think I was, but for some reason the router and by extension, our script, was misreporting this for some reason.
> > ...
> > So the router was probably running in a state lower than 80MB but for some reason it was always reporting 80MB free.
> 
> If the router's kernel is indeed reporting the wrong information for "Available RAM" that would be a major bug since other processes usually rely on information from "**/proc/meminfo**" to evaluate various states of current memory.
> 
> Anyway, have a good night's rest buddy. Looks like you certainly had a full day today running those tests back & forth and collecting data points to figure out what was happening. I admire your persistence & determination. My hat's off to you, brother!!

I promise you no matter what I did, I could not get it lower than 80MB (78MB to be exact when it finally started properly reporting).

Which is why I kinda want to retest on my primary, to see if it's a real bug, or something wrong with my specific node.

If my primary does go below 80MB then I can exclude point number 2 from my reasons above. And by extension, probably point 1.

Either way, I may poke at that more tomorrow and let you know. It should be fairly easy to test considering I won't do the actual flash.

Goodnight buddy, sorry for the spam. Lol it's more like note taking and documentation as we go that we can reference back too.

Martinski4GitHub:
> And to wrap up the thought, this is the 2 reasons why I removed my suggestion for the memory management from my PR.
> 
>     1. I wasn't able to get a successful flash without being in the 140MB free range which is more in line with our current 50% overhead requirements.
> 
>     2. Seeing this potential issue with routers miss reporting below 80MB doesn't give me the confidence that all routers could use that adjustment logic correctly with the 40MB cushion we originally intended.
> 
> 
> If we ever re-implement such logic, it would again be with a 40MB cushion for safety, and only for the 512MB models. For now we would need a bigger sample size, I tried. Trust me I did. But I can't replicate correctly so if another user has an issue down the line we can try and grow the sample size that way.
>
Yeah, I understand and I agree with your assessment. We need more data points from different router models with 512MB RAM that experience the "low memory" condition. OOH, I hope those are very rare cases (I don't think my router & my cousin's router are the outliers); OTOH, we need more such cases to get a clear picture of what's causing those routers to lose so much RAM that they don't get back sufficient "Available RAM" to continue with the F/W flash.

In any case, the "better safe than sorry" approach is a good thing here, especially when flashing F/W images.

ExtremeFiretop:
@Martinski4GitHub 

Today I retested on my Primary router, here are the results:

Primary Router Tests:

151MB (Zip Size) + -75%% (Overhead) = 37.5MB (Expected Requirement)

To Identify Available Memory...
Used : theMemAvailable_KB="$(cat /proc/meminfo | awk -F ' ' '/^MemAvailable: /{print $2}')"
And : echo "$theMemAvailable_KB"
Between each run...

Basic Math:
520.5MB (Available) - 160MB (Expected usage) = 360.5MB (To use up)
MINUS (-) 35MB (To leave available) = 325.5MB (Total junk to fill)

1st Pass:
185 MB (Available After junk) = RAM Decreased to 66MB. (Stopped Flash)

Increased Junk Size 20MB

2nd Pass:
181 MB (Available After junk) = RAM Decreased to 66MB. (Stopped Flash)

So it got lower than 78MB, but it seems to stop at 66MB even though I decreased it by 20MB still.
Seems to be the same general issue but not the same values.

ExtremeFiretop:
@Martinski4GitHub 

We have some kinda bug it seems:

========= By ExtremeFiretop & Martinski W. ==========

-----------------------------------------------------
      F/W Product/Model ID:  GT-AXE11000
      F/W Update Available:  NONE FOUND
      F/W Version Installed: 3004.388.6.2_rog
      USB Storage Connected: True
-----------------------------------------------------
  1.  Run F/W Update Check Now

  2.  Configure Router Login Credentials

  3.  Toggle F/W Update Check
      [Currently DISABLED]
      [Last Notification Date: 2024-Mar-24 10:39 PM]

  4.  Set F/W Update Postponement Days
      [Current Days: 0]

  5.  Set F/W Update Check Schedule
      [Current Schedule: 15 4 * * *]

 em.  Toggle F/W Update Email Notifications
      [Currently ENABLED, Format: HTML]

 ad.  Advanced Options

 un.  Uninstall

  e.  Exit
-----------------------------------------------------
Enter selection:  1

MerlinAU v1.0.9
Running the update task now... Checking for F/W updates...
Firmware update check is currently disabled.

Press <Enter> to continue...
**/jffs/scripts/MerlinAU.sh: line 17: 2: parameter not set**

That was on my router..
This is on my side gigs router:

image

Same thing the user reported in the forums.

ExtremeFiretop:
Now I can't recreate the issue.
I was only able too after I downloaded our latest update, after I closed it and re-opened it the issue was gone.

ExtremeFiretop:
Maybe we try making cru l the default first attempt instead of crontab?

```
if [ -z "$(which cru)" ]; then
    cronListCmd="crontab -l"
else
    cronListCmd="cru l"
fi
```

ExtremeFiretop:
We are using    cru a "$CRON_JOB_TAG" "$newSchedule $CRON_JOB_RUN"

To add the cron so maybe we should also default to cru for the list, that was my train of thought, also all the errors seem to complain about the command crontab -l but not cru l.

Martinski4GitHub:
> 
> 
> em. Toggle F/W Update Email Notifications [Currently ENABLED, Format: HTML]
> 
> ad. Advanced Options
> 
> un. Uninstall
> 
> e. Exit
> 
> Enter selection: 1
> 
> MerlinAU v1.0.9 Running the update task now... Checking for F/W updates... Firmware update check is currently disabled.
> 
> Press to continue... **/jffs/scripts/MerlinAU.sh: line 17: 2: parameter not set**
> 
I suspect the issue might have been related to this statement:
`if [ "$1" = "**ERROR**" ] && [ "$2" = "**NO_URL**" ]`

Which I've now changed to:
`if [ "$release_version" = "**ERROR**" ] && [ "$release_link" = "**NO_URL**" ]`

I also moved up that check to be right below where the values were obtained & assigned because we need to do that check as soon as we get the values (see PR #176).

Martinski4GitHub:
> Now I can't recreate the issue. I was only able too after I downloaded our latest update, after I closed it and re-opened it the issue was gone.
>
Which one specifically? You reported 2 separate problems.

Martinski4GitHub:
> Maybe we try making cru l the default first attempt instead of crontab?
> 
> ```
> if [ -z "$(which cru)" ]; then
>     cronListCmd="crontab -l"
> else
>     cronListCmd="cru l"
> fi
> ```
The problem appears to be how the variable "$cronListCmd" is evaluated when referenced. Apparently, in some router environments, sometimes the variable is **not** evaluated as containing a command but as a referenced quoted string. So I changed each call to make sure each time it's forced to be evaluated as a command (see PR #176).

Since I could not recreate the problem on my router, please try to check if the latest changes worked in your router that experienced the problem.

Thanks.

ExtremeFiretop:
> > Now I can't recreate the issue. I was only able too after I downloaded our latest update, after I closed it and re-opened it the issue was gone.
> 
> Which one specifically? You reported 2 separate problems.

I only reported the one really, the primary router was just to validate the issues from yesterday. :)
But yes specifically the issue with the crontab, it disappeared after I stopped and started the script so will be hard to test this part:

> Since I could not recreate the problem on my router, please try to check if the latest changes worked in your router that experienced the problem.
> 
> Thanks.

I'll review the PR and your train of thought later :)

ExtremeFiretop:
> > Maybe we try making cru l the default first attempt instead of crontab?
> > ```
> > if [ -z "$(which cru)" ]; then
> >     cronListCmd="crontab -l"
> > else
> >     cronListCmd="cru l"
> > fi
> > ```
> 
> The problem appears to be how the variable "$cronListCmd" is evaluated when referenced. Apparently, in some router environments, sometimes the variable is **not** evaluated as containing a command but as a referenced quoted string. So I changed each call to make sure each time it's forced to be evaluated as a command (see PR #176).
> 
> Since I could not recreate the problem on my router, please try to check if the latest changes worked in your router that experienced the problem.
> 
> Thanks.

I have zero idea how you got there.
I see the error: **printf "${REDct}**ERROR**${NOct}: Failed to add the cron job [${CRON_JOB_TAG}].\n"**

I look in the code for that and find the if statement:

```
       if _AddCronJobEntry_
       then
           printf "Cron job '${GRNct}${CRON_JOB_TAG}${NOct}' was added successfully.\n"
           _AddCronJobRunScriptHook_
       else
           printf "${REDct}**ERROR**${NOct}: Failed to add the cron job [${CRON_JOB_TAG}].\n"
       fi
```

I review the **_AddCronJobEntry_** function that called it.
Inside the function, I clearly see:

```
   cru a "$CRON_JOB_TAG" "$newSchedule $CRON_JOB_RUN"
   sleep 1
   if $cronListCmd | grep -qE "$CRON_JOB_RUN #${CRON_JOB_TAG}#$"
```

So it calls the cronListCmd commandlet we created, looking up the cronListCmd I see command for crontab -l:

```
if [ -z "$(which cru)" ]; then
    cronListCmd="crontab -l"
else
    cronListCmd="cru l"
fi
```

Which is why I figured you made the change you did yesterday to that format, from the old format:

```
cronCmd="$(which crontab) -l"
[ "$cronCmd" = " -l" ] && cronCmd="$(which cru) l"
```

How you took away that this line is the problem:

```
    if [ "$1" = "**ERROR**" ] && [ "$2" = "**NO_URL**" ]
    then
        Say "${REDct}**ERROR**${NOct}: No firmware release URL was found for [$PRODUCT_ID] router model."
        "$inMenuMode" && _WaitForEnterKey_ "$mainMenuReturnPromptStr"
        return 1
    fi
```

I have zero idea, I would assume to see "Say "**${REDct}**ERROR**${NOct}: No firmware release URL was found for [$PRODUCT_ID] router model.""**

Considering that's the if statement for if any of those aren't populated.
My default thought would be that 

```
    # Use set to read the output of the function into variables
    set -- $(_GetLatestFWUpdateVersionFromWebsite_ "$FW_URL_RELEASE")
    release_version="$1"
    release_link="$2"
```

Is working fine since we use release_version="$1" right below it in: 

```
    if ! _CheckTimeToUpdateFirmware_ "$current_version" "$release_version"
    then
        "$inMenuMode" && _WaitForEnterKey_ "$mainMenuReturnPromptStr"
        return 0
    fi
```

And use release_link="$2" right here:

```
    # Get the required space for the firmware download and extraction
    required_space_kb=$(get_required_space "$release_link")
    if ! _HasRouterMoreThan256MBtotalRAM_ && [ "$required_space_kb" -gt 51200 ]; then
        if ! _ValidateUSBMountPoint_ "$FW_ZIP_BASE_DIR" 1
        then
            Say "${REDct}**ERROR**${NOct}: A USB drive is required for the F/W update due to limited RAM."
            "$inMenuMode" && _WaitForEnterKey_ "$mainMenuReturnPromptStr"
            return 1
        fi
    fi
```

I'd think one of those would complain if we had a problem with any of those values, right?
Maybe you can explain your train of thought so I can learn and take away how the heck you made this connection?

ExtremeFiretop:
> > Which one specifically? You reported 2 separate problems.
> 
> I only reported the one really, the primary router was just to validate the test results from yesterday. :) 

Oh I see what you mean by I reported too. You mean by the screenshots the 2 different errors between my router and my side gigs router.

Tbh, I expected those to be interrelated, but you could be right and they aren't. But again I'd need to know how you made this connection cause to me this is dark magic lol!

PR #171, opened 2024-03-24
Toggle_Patch

ExtremeFiretop:
Toggle_Patch

Added additional code to disable the BACKUPMON toggle if not installed.
Also, the backupmon toggle shouldn't be a number because it won't always be displayed.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4118,17 +4118,6 @@ _ShowAdvancedOptionsMenu_()
        printf "\n${padStr}[Currently ${GRNct}ENABLED${NOct}]\n"
    fi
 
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
> 
> Added additional code to disable the BACKUPMON toggle if not installed. Also, the backupmon toggle shouldn't be a number because it won't always be displayed.
>
It looks like we both found the same problem around the menu setup for the toggle but fixed it differently.
I decided to make the 2 menu options that are not always visible non-numeric so that anytime we add a new menu selection the non-numeric selection is no longer an issue.

ExtremeFiretop:
Yours fixes more so yours gets merged :)

PR #180, opened 2024-03-27
Log memory stats when trying to free memory for F/W flash.

Martinski4GitHub:
Added code to log memory stats at the points where we're trying to free memory to be able to continue with the F/W flash.


ExtremeFiretop (approved):

ExtremeFiretop:
That's an instant approve. No need to second guess on that one!

I see what you mean, we will get a breakdown of all the key proc values on every attempt of the cleanup up until the 3rd attempt.

I like it, at least next time if someone has the issue we can clearly point at something in the logs. (Hopefully!)

PR #191, opened 2024-04-14
Disable Email Notifications for Available Updates from Node

ExtremeFiretop:
Disable Email Notifications for Available Updates from Node
Primary now handles these updates notifications for the nodes.

Even if they want notifications enabled on the node for flashing, unzip errors, ect, they shouldn't be receiving the update available notifications.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -3043,7 +3043,10 @@ _CheckNewUpdateFirmwareNotification_()
            fwNewUpdateNotificationDate="$(date +"$FW_UpdateNotificationDateFormat")"
            Update_Custom_Settings FW_New_Update_Notification_Vers "$fwNewUpdateNotificationVers"
            Update_Custom_Settings FW_New_Update_Notification_Date "$fwNewUpdateNotificationDate"
-           _SendEMailNotification_ NEW_FW_UPDATE_STATUS
+           if "$inRouterSWmode" 
+           then
+              _SendEMailNotification_ NEW_FW_UPDATE_STATUS
+           fi
        fi
    fi
 
@@ -3052,7 +3055,10 @@ _CheckNewUpdateFirmwareNotification_()
    then
        fwNewUpdateNotificationDate="$(date +"$FW_UpdateNotificationDateFormat")"
        Update_Custom_Settings FW_New_Update_Notification_Date "$fwNewUpdateNotificationDate"
-       _SendEMailNotification_ NEW_FW_UPDATE_STATUS
+       if "$inRouterSWmode" 
+       then
+          _SendEMailNotification_ NEW_FW_UPDATE_STATUS
+       fi
    fi
    return 0
 }
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -8,7 +8,7 @@
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION=1.1.0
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
> Disable Email Notifications for Available Updates from Node Primary now handles these updates notifications for the nodes.
> 
> Even if they want notifications enabled on the node for flashing, unzip errors, ect, they shouldn't be receiving the update available notifications.
> 
That's a good point. But now I'm wondering if that's something that should be a user-configurable option in "Advanced Settings" (shown **only** for Mesh Nodes, with **FALSE** as the default) because I'd bet that some user(s) out there will ask to be able to receive email notifications for everything for some reason (e.g. the scheduled cron job runs at different times with shorter interval than the Main router so the update notifications would be received sooner). Just a thought.

ExtremeFiretop:
> > Disable Email Notifications for Available Updates from Node Primary now handles these updates notifications for the nodes.
> > Even if they want notifications enabled on the node for flashing, unzip errors, ect, they shouldn't be receiving the update available notifications.
> 
> That's a good point. But now I'm wondering if that's something that should be a user-configurable option in "Advanced Settings" (shown **only** for Mesh Nodes, with **FALSE** as the default) because I'd bet that some user(s) out there will ask to be able to receive email notifications for everything for some reason (e.g. the scheduled cron job runs at different times with shorter interval than the Main router so the update notifications would be received sooner). Just a thought.

I had thought about this already. but I couldn't see a reason or use-case myself so I decided to opt for this.
It would add another option in the advanced options which would likely be unused, I'd probably wait until someone asks for this as a feature myself and then re-assess based on why they might be requesting this in the first place.

ExtremeFiretop:
> > Disable Email Notifications for Available Updates from Node Primary now handles these updates notifications for the nodes.
> > Even if they want notifications enabled on the node for flashing, unzip errors, ect, they shouldn't be receiving the update available notifications.
> 
> That's a good point. But now I'm wondering if that's something that should be a user-configurable option in "Advanced Settings" (shown **only** for Mesh Nodes, with **FALSE** as the default) because I'd bet that some user(s) out there will ask to be able to receive email notifications for everything for some reason (e.g. the scheduled cron job runs at different times with shorter interval than the Main router so the update notifications would be received sooner). Just a thought.

I guess the way I saw it was the primary reporting the status of the nodes wasn't made optional, so the nodes reporting themselves probably shouldn't be made optional, unless we made it all optional.

Martinski4GitHub:
> > > Disable Email Notifications for Available Updates from Node Primary now handles these updates notifications for the nodes.
> > > Even if they want notifications enabled on the node for flashing, unzip errors, ect, they shouldn't be receiving the update available notifications.
> > 
> > 
> > That's a good point. But now I'm wondering if that's something that should be a user-configurable option in "Advanced Settings" (shown **only** for Mesh Nodes, with **FALSE** as the default) because I'd bet that some user(s) out there will ask to be able to receive email notifications for everything for some reason (e.g. the scheduled cron job runs at different times with shorter interval than the Main router so the update notifications would be received sooner). Just a thought.
> 
> I guess the way I saw it was the primary reporting the status of the nodes wasn't made optional, so the nodes reporting themselves probably shouldn't be made optional, unless we made it all optional.
>
OK, that sounds good. Let's wait for a user request to make the Mesh Nodes notifications user-configurable.

PR #197, opened 2024-04-27
Bug Fix for Node Email Notifications

ExtremeFiretop:
Bug Fix for Node Email Notifications
I found some bugs with the node feature with this latest firmware release btw that are in our current production version:

1. Even though the node has a name, the email says TBD

image

image

2. I'm not sure if the email title is accurate, in my case the nodes are GT-AXE11000's but I don't remember making that dynamic. Will need to review.

image

3. I am able to trigger the email every time I go to the nodes menu.

PR #201, opened 2024-04-29
Node_MinorLogicFix

ExtremeFiretop:
Node_MinorLogicFix

As mentioned in PR: 197 here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/197/files/d2f86af7c833992d0593a1a51166c95c2d203150

So about the frozen cron I wanted to discuss with you.
Originally, I didn't understand why moving the code from the email function resolved the issue.

All that made sense to me is somehow something wasn't being run/processed/populated when run as a cron job. Now going back through the code, I notice that:

`node_online_status=$(_NodeActiveStatus_)`

Is only located under the: `_ShowNodesMenu_` code.
But it looks like required code for the `_ProcessMeshNodes_` to run.

I'm assuming something like this is causing us our issues.
Any recommendations on where to move it? I'm thinking right below the:

`if ! node_list="$(_GetNodeIPv4List_)" then node_list="" ; fi`

of the `_ProcessMeshNodes_` function.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4162,24 +4162,6 @@ _SetEMailFormatType_()
    _WaitForEnterKey_ "$advnMenuReturnPromptStr"
 }
 
-# RegExp for IPv4 address #
-readonly IPv4octet_RegEx="([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-5][0-5])"
-readonly IPv4addrs_RegEx="((${IPv4octet_RegEx}\.){3}${IPv4octet_RegEx})"
-readonly IPv4privt_RegEx="((^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.))"
-
-##----------------------------------------##
-## Modified by Martinski W. [2024-Apr-06] ##
-##----------------------------------------##
-_ValidatePrivateIPv4Address_()
-{
-   if [ $# -eq 0 ] || [ -z "$1" ] || \
-      ! echo "$1" | grep -qE "^${IPv4addrs_RegEx}$" || \
-      ! echo "$1" | grep -qE "^${IPv4privt_RegEx}"
-   then return 1
-   else return 0
-   fi
-}
-
 _ProcessMeshNodes_()
 {
     includeExtraLogic="$1"  # Use '1' to include extra logic, '0' to exclude
@@ -4768,6 +4750,70 @@ _ShowAdvancedOptionsMenu_()
 
    printf "\n  ${GRNct}e${NOct}.  Return to Main Menu\n"
    printf "${SEPstr}"
+}
+
+# RegExp for IPv4 address #
+readonly IPv4octet_RegEx="([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-5][0-5])"
+readonly IPv4addrs_RegEx="((${IPv4octet_RegEx}\.){3}${IPv4octet_RegEx})"
+readonly IPv4privt_RegEx="((^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.))"
+
+##----------------------------------------##
+## Modified by Martinski W. [2024-Apr-06] ##
+##----------------------------------------##
+_ValidatePrivateIPv4Address_()
+{
+   if [ $# -eq 0 ] || [ -z "$1" ] || \
+      ! echo "$1" | grep -qE "^${IPv4addrs_RegEx}$" || \
+      ! echo "$1" | grep -qE "^${IPv4privt_RegEx}"
+   then return 1
+   else return 0
+   fi
+}
+
+##----------------------------------------##
+## Modified by Martinski W. [2024-Apr-06] ##
+##----------------------------------------##
+_ProcessMeshNodes_()
+{
+    includeExtraLogic="$1"  # Use '1' to include extra logic, '0' to exclude
+    if [ $# -eq 0 ] || [ -z "$1" ]
+    then echo "**ERROR** **NO_PARAMS**" ; return 1 ; fi
+
+    uid=1
+    if ! node_list="$(_GetNodeIPv4List_)"
+    then node_list="" ; fi
+
+    if "$inRouterSWmode"
+    then
+        if [ -n "$node_list" ]
+        then
+            # Iterate over the list of nodes and print information for each node
+            for nodeIPv4addr in $node_list
+            do
+                ! _ValidatePrivateIPv4Address_ "$nodeIPv4addr" && continue
+                _GetNodeInfo_ "$nodeIPv4addr"
+                if ! Node_FW_NewUpdateVersion="$(_GetLatestFWUpdateVersionFromNode_ 1)"
+                then
+                    Node_FW_NewUpdateVersion="NONE FOUND"
+                else
+                    _CheckNodeFWUpdateNotification_ "$Node_combinedVer" "$Node_FW_NewUpdateVersion"
+                fi
+
+                # Apply extra logic if flag is '1'
+                if [ "$includeExtraLogic" -eq 1 ]; then
+                    _PrintNodeInfo "$nodeIPv4addr" "$node_online_status" "$Node_FW_NewUpdateVersion" "$uid"
+                    uid="$((uid + 1))"
+                fi
+            done
+            if [ -s "$tempNodeEMailList" ]; then
+                _SendEMailNotification_ AGGREGATED_UPDATE_NOTIFICATION
+            fi
+        else
+            if [ "$includeExtraLogic" -eq 1 ]; then
+                printf "\n${padStr}${padStr}${padStr}${REDct}No AiMesh Node(s)${NOct}"
+            fi
+        fi
+    fi
 }
 
 ##---------------------------------------##
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4162,6 +4162,24 @@ _SetEMailFormatType_()
    _WaitForEnterKey_ "$advnMenuReturnPromptStr"
 }
 
+# RegExp for IPv4 address #
+readonly IPv4octet_RegEx="([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-5][0-5])"
+readonly IPv4addrs_RegEx="((${IPv4octet_RegEx}\.){3}${IPv4octet_RegEx})"
+readonly IPv4privt_RegEx="((^10\.)|(^172\.1[6-9]\.)|(^172\.2[0-9]\.)|(^172\.3[0-1]\.)|(^192\.168\.))"
+
+##----------------------------------------##
+## Modified by Martinski W. [2024-Apr-06] ##
+##----------------------------------------##
+_ValidatePrivateIPv4Address_()
+{
+   if [ $# -eq 0 ] || [ -z "$1" ] || \
+      ! echo "$1" | grep -qE "^${IPv4addrs_RegEx}$" || \
+      ! echo "$1" | grep -qE "^${IPv4privt_RegEx}"
+   then return 1
+   else return 0
+   fi
+}
+
 _ProcessMeshNodes_()
 {
     includeExtraLogic="$1"  # Use '1' to include extra logic, '0' to exclude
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
> Node_MinorLogicFix
> 
> As mentioned in PR: 197 here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/197/files/d2f86af7c833992d0593a1a51166c95c2d203150
> 
> So about the frozen cron I wanted to discuss with you. Originally, I didn't understand why moving the code from the email function resolved the issue.
> 
> All that made sense to me is somehow something wasn't being run/processed/populated when run as a cron job. Now going back through the code, I notice that:
> 
> `node_online_status=$(_NodeActiveStatus_)`
> 
> Is only located under the: `_ShowNodesMenu_` code. But it looks like required code for the `_ProcessMeshNodes_` to run.
> 
The "`_ProcessMeshNodes_`" function already calls "`_GetNodeIPv4List_`" to get the list of current mesh nodes, what would calling "`_NodeActiveStatus_`" give you that "`_GetNodeIPv4List_`" does not already provide?

> I'm assuming something like this is causing us our issues. Any recommendations on where to move it? I'm thinking right below the:
> 
> `if ! node_list="$(_GetNodeIPv4List_)" then node_list="" ; fi`
> 
> of the `_ProcessMeshNodes_` function.
>
That doesn't look correct to me, based on my understanding of the "`_NodeActiveStatus_`" function, but I don't have mesh nodes to fully verify.

ExtremeFiretop:
> > Node_MinorLogicFix
> > As mentioned in PR: 197 here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/197/files/d2f86af7c833992d0593a1a51166c95c2d203150
> > So about the frozen cron I wanted to discuss with you. Originally, I didn't understand why moving the code from the email function resolved the issue.
> > All that made sense to me is somehow something wasn't being run/processed/populated when run as a cron job. Now going back through the code, I notice that:
> > `node_online_status=$(_NodeActiveStatus_)`
> > Is only located under the: `_ShowNodesMenu_` code. But it looks like required code for the `_ProcessMeshNodes_` to run.
> 
> The "`_ProcessMeshNodes_`" function already calls "`_GetNodeIPv4List_`" to get the list of current mesh nodes, what would calling "`_NodeActiveStatus_`" give you that "`_GetNodeIPv4List_`" does not already provide?
> 
> > I'm assuming something like this is causing us our issues. Any recommendations on where to move it? I'm thinking right below the:
> > `if ! node_list="$(_GetNodeIPv4List_)" then node_list="" ; fi`
> > of the `_ProcessMeshNodes_` function.
> 
> That doesn't look correct to me, based on my understanding of the "`_NodeActiveStatus_`" function, but I don't have mesh nodes to fully verify.

The _GetNodeIPv4List_ is only to get the "Perminante" list.
No matter if a node is online, or offline, the asus_device_list nvram value will always be populated and always return an IP address of the node.

The primary remembers this device in this value.

The _NodeActiveStatus_ is actually a live nvram value, it only shows when nodes are alive and communications now at this moment. 

With this information, If cfg_device_list returns no IP of any nodes, but asus_device_list returns an IP of the node, it means we have a node setup, but it's offline for some reason. They actually don't have the same purpose and work differently.

How would the _ProcessMeshNodes_ function run without this?  When it's a required here?

image

ExtremeFiretop:
> > > Node_MinorLogicFix
> > > As mentioned in PR: 197 here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/197/files/d2f86af7c833992d0593a1a51166c95c2d203150
> > > So about the frozen cron I wanted to discuss with you. Originally, I didn't understand why moving the code from the email function resolved the issue.
> > > All that made sense to me is somehow something wasn't being run/processed/populated when run as a cron job. Now going back through the code, I notice that:
> > > `node_online_status=$(_NodeActiveStatus_)`
> > > Is only located under the: `_ShowNodesMenu_` code. But it looks like required code for the `_ProcessMeshNodes_` to run.
> > 
> > 
> > The "`_ProcessMeshNodes_`" function already calls "`_GetNodeIPv4List_`" to get the list of current mesh nodes, what would calling "`_NodeActiveStatus_`" give you that "`_GetNodeIPv4List_`" does not already provide?
> > > I'm assuming something like this is causing us our issues. Any recommendations on where to move it? I'm thinking right below the:
> > > `if ! node_list="$(_GetNodeIPv4List_)" then node_list="" ; fi`
> > > of the `_ProcessMeshNodes_` function.
> > 
> > 
> > That doesn't look correct to me, based on my understanding of the "`_NodeActiveStatus_`" function, but I don't have mesh nodes to fully verify.
> 
> The _GetNodeIPv4List_ is only to get the "Permanent" list. No matter if a node is online, or offline, the asus_device_list nvram value will always be populated and always return an IP address of the node.
> 
> The primary remembers this device in this value.
> 
> The _NodeActiveStatus_ is actually a live nvram value, it only shows when nodes are alive and communications now at this moment.
> 
> With this information, If cfg_device_list returns no IP of any nodes, but asus_device_list returns an IP of the node, it means we have a node setup, but it's offline for some reason. They actually don't have the same purpose and work differently.
> 
> How would the _ProcessMeshNodes_ function run without this? When it's a required here?
> 
> image
> 
> **Edit:**
> 
> NVM, I need some coffee this morning, $node_online_status is not required as a CRON* But it is required when INTERACTIVE*
> 
> I forgot the print node function info is only for the menu. Not a cron.

ExtremeFiretop:
Investigation continues...

ExtremeFiretop:
Fixed it:

image

ExtremeFiretop:
Okay @Martinski4GitHub I'm confident this is a fix, it was a placement issue for the ProcessNodes function and nothing else :)
I'll submit a similar change for the Gnuton branch next.

ExtremeFiretop:
Okay buddy

As discussed in the forums with Viktor:

I think one of the easiest things to start is change the current defaults post installation. 
This setup isn't ideal for a newbie user that doesn't understand all the features and switches yet.
I'm cool with leaving a default postpone, but maybe we change the default cron to daily as well would be in our benefit.

Martinski4GitHub:
> Okay @Martinski4GitHub I'm confident this is a fix, it was a placement issue for the ProcessNodes function and nothing else with the code (duh) haha :) I'll submit a similar change for the Gnuton branch next.
>
Ah, yes, correct placement of the function; otherwise, it's not yet defined when running in non-interactive mode. That makes sense now. Good catch!!

Martinski4GitHub:
> Okay buddy
> 
> As discussed in the forums with Viktor:
> 
> I think one of the easiest things to start is change the current defaults post installation. This setup isn't ideal for a newbie user that doesn't understand all the features and switches yet. I'm cool with leaving a default postpone, but maybe we change the default cron to daily as well would be in our benefit.
>
Yeah, OK. A daily cron job as default with the 7-day postponement sounds good.

Martinski4GitHub:
> Okay buddy
> 
> As discussed in the forums with Viktor:
> 
> I think one of the easiest things to start is change the current defaults post installation. This setup isn't ideal for a newbie user that doesn't understand all the features and switches yet. I'm cool with leaving a default postpone, but maybe we change the default cron to daily as well would be in our benefit.
>
Now, I'm thinking that when running on a mesh node, the default cron job schedule should be set 15 minutes **after** the primary router just to make sure there are no conflicts WRT going online to check & download the F/W zip files (you alluded to this in an SNB Forum post).

For instance, "`0  0  *  *  *`" for primary router, and "`15  0  *  *  *`" for all mesh nodes. This would prevent possible issues that may occur, especially when some users have several mesh nodes installed.

What do you think?

ExtremeFiretop:
> > Okay buddy
> > As discussed in the forums with Viktor:
> > I think one of the easiest things to start is change the current defaults post installation. This setup isn't ideal for a newbie user that doesn't understand all the features and switches yet. I'm cool with leaving a default postpone, but maybe we change the default cron to daily as well would be in our benefit.
> 
> Now, I'm thinking that when running on a mesh node, the default cron job schedule should be set 15 minutes **after** the primary router just to make sure there are no conflicts WRT going online to check & download the F/W zip files (you alluded to this in an SNB Forum post).
> 
> For instance, "`0 0 * * *`" for primary router, and "`15 0 * * *`" for all mesh nodes. This would prevent possible issues that may occur, especially when some users have several mesh nodes installed.
> 
> What do you think?

Yes I actually do that with my setup, I do the nodes first since the nodes aren't as important as the primary, if something goes wrong I rather the node goes down first.

But that's a personal opinion, on the other side of that coin is if the primary goes down first and there a real problem the nodes probably didn't get the update after that.

Depends how much you trust the updates and how much you might monitor them.

But really I think the key point of having the 15 minute offset to avoid conflict is smart.

ExtremeFiretop:
> > Okay @Martinski4GitHub I'm confident this is a fix, it was a placement issue for the ProcessNodes function and nothing else with the code (duh) haha :) I'll submit a similar change for the Gnuton branch next.
> 
> Ah, yes, correct placement of the function; otherwise, it's not yet defined when running in non-interactive mode. That makes sense now. Good catch!!

Thank you! I was confused this morning because I was sure I had found the cause but I didn't actually dig into what the value was for I just noticed a value in the function that wouldn't be populated. But then realized it shouldn't be when run as a cron so I had to go back to the drawing board.

Are we approving this PR or did you include these changes in your latest PR?

I'm out now but about to be back home in some 20 minutes or so to have dinner and review.

Martinski4GitHub:
> > Ah, yes, correct placement of the function; otherwise, it's not yet defined when running in non-interactive mode. That makes sense now. Good catch!!
> 
> Thank you! I was confused this morning because I was sure I had found the cause but I didn't actually dig into what the value was for I just noticed a value in the function that wouldn't be populated. But then realized it shouldn't be when run as a cron so I had to go back to the drawing board.
> 
> Are we approving this PR or did you include these changes in your latest PR?
> 
No, I didn't include/merge the changes in my PR. I was waiting for you to respond and see if you wanted to implement a separate default cron job schedule for the nodes. IOW, set up to separate default schedules: one for primary and another for nodes. But thinking about it just now, that should really be a separate PR anyway.

Martinski4GitHub (approved):
Good to go.

Martinski4GitHub:
> > 
> > Now, I'm thinking that when running on a mesh node, the default cron job schedule should be set 15 minutes **after** the primary router just to make sure there are no conflicts WRT going online to check & download the F/W zip files (you alluded to this in an SNB Forum post).
> > For instance, "`0 0 * * *`" for primary router, and "`15 0 * * *`" for all mesh nodes. This would prevent possible issues that may occur, especially when some users have several mesh nodes installed.
> > What do you think?
> 
> Yes I actually do that with my setup, I do the nodes first since the nodes aren't as important as the primary, if something goes wrong I rather the node goes down first.
> 
> But that's a personal opinion, on the other side of that coin is if the primary goes down first and there a real problem the nodes probably didn't get the update after that.
> 
> Depends how much you trust the updates and how much you might monitor them.
> 
Yes, that's why the 15-day postponement as default should remain at least. It gives users 2 weeks to check & review the changelog independently and monitor any feedback in the forums before allowing the F/W update. In fact, now with a daily cron job as default, I'm wondering if we should increase the default delay to 21 days to give users more than enough time before "taking the plunge." I always like to wait & read as much feedback as I can before doing the F/W Update because I need a consistent & reliable network connection so I'm always wary of anything that might disrupt that.

ExtremeFiretop:
> > > Now, I'm thinking that when running on a mesh node, the default cron job schedule should be set 15 minutes **after** the primary router just to make sure there are no conflicts WRT going online to check & download the F/W zip files (you alluded to this in an SNB Forum post).
> > > For instance, "`0 0 * * *`" for primary router, and "`15 0 * * *`" for all mesh nodes. This would prevent possible issues that may occur, especially when some users have several mesh nodes installed.
> > > What do you think?
> > 
> > 
> > Yes I actually do that with my setup, I do the nodes first since the nodes aren't as important as the primary, if something goes wrong I rather the node goes down first.
> > But that's a personal opinion, on the other side of that coin is if the primary goes down first and there a real problem the nodes probably didn't get the update after that.
> > Depends how much you trust the updates and how much you might monitor them.
> 
> Yes, that's why the 15-day postponement as default should remain at least. It gives users 2 weeks to check & review the changelog independently and monitor any feedback in the forums before allowing the F/W update. In fact, now with a daily cron job as default, I'm wondering if we should increase the default delay to 21 days to give users more than enough time before "taking the plunge." I always like to wait & read as much feedback as I can before doing the F/W Update because I need a consistent & reliable network connection so I'm always wary of anything that might disrupt that.

Send it!

PR #202, opened 2024-05-01
Code improvements.

Martinski4GitHub:
Code improvements in the "`_GetAllNodeSettings_` " function.

PR #204, opened 2024-05-03
Include New Logs Menu

ExtremeFiretop:
Include New Logs Menu

Hey @Martinski4GitHub Created PR: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/204
As discussed in PR: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/202

> I'll let you take the lead on this feature and decide what works best from a user's perspective. My own perspective is completely different because I have a more hands-on, technical-oriented approach to how I review the Changelog & monitor user feedback before I decide when to flash the latest F/W image. But as mentioned, I'm OK with whatever you decide.

**This includes the suggestion you made below:**

> If I understand the desired end-result, wouldn't it be simpler to just provide a menu option to download the appropriate Changelog "on demand" **without** actually downloading any F/W ZIP file?
> 
> For example, we add a new menu option like: `cl) View the latest F/W Changelog file.`
> 
> When the user enters "**cl**" the script would download the appropriate data (via **curl** cmd) using the corresponding URL.

**As the agreed upon method:**

> The more I think about it, the more I like this solution even though it isn't simpler, why? because it's available all the time, even if the user hasn't setup email notifications. That removes the dependency to force people to set that up to review the changelogs.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -3136,17 +3136,16 @@ _CheckNodeFWUpdateNotification_()
    return 0
 }
 
-##------------------------------------------##
-## Modified by ExtremeFiretop [2024-May-03] ##
-##------------------------------------------##
+##----------------------------------------##
+## Modified by Martinski W. [2024-Mar-21] ##
+##----------------------------------------##
 _CheckTimeToUpdateFirmware_()
 {
    if [ $# -lt 2 ] || [ -z "$1" ] || [ -z "$2" ]
    then echo "**ERROR** **NO_PARAMS**" ; return 1 ; fi
 
-   local notifyTimeSecs postponeTimeSecs currentTimeSecs dstAdjustSecs dstAdjustDays
-   local fwNewUpdateNotificationDate fwNewUpdateNotificationVers fwNewUpdatePostponementDays
-   local currCronSchedule cronHour cronMinute nextCronRunSecs
+   local notifyTimeSecs  postponeTimeSecs  currentTimeSecs  dstAdjustSecs  dstAdjustDays
+   local fwNewUpdateNotificationDate  fwNewUpdateNotificationVers  fwNewUpdatePostponementDays
 
    _CheckNewUpdateFirmwareNotification_ "$1" "$2"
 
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
@Martinski4GitHub buddy!

While I said I wanted to avoid a specific logs menu if possible, I don't know a better way to make it look.
My advanced options is getting long, and so is main menu, and I can't put it under mesh nodes for obvious reasons.

That left me with creating a new menu, but I guess you can use this new menu for your MerlinAU logs selection option :)

MerlinAU.sh
```diff
@@ -1272,6 +1279,8 @@ _CreateEMailContent_()
            emailBodyTitle="WARNING"
            {
              echo "Found high-risk phrases in the change-logs while Auto-Updating to version <b>${fwNewUpdateVersion}</b> on the <b>${MODEL_ID}</b> router."
+             echo "Changelog contents include the following changes:"
```
ExtremeFiretop, line 1282: I decided to go with both methods, have it in the email, and have it available on demand :)
Why not.

ExtremeFiretop:
Added the 15 minute offset for the nodes in commit:

https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/204/commits/cfc9f41313c413acfd38143212deabec37ad62dc

ExtremeFiretop:
@Martinski4GitHub 

Modified the time calculations to factor in the postpone time and cron time in commit:
https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/204/commits/2dfb984d2801a6b01a3542810dd076d8e293a425

As suggested by Viktor in the forums...
Here are some examples;

**POSTPONE 7 DAYS and 4AM CRON:**
image

**POSTPONE 7 DAYS and 6AM CRON:**
image

**POSTPONE 1 DAY and 6AM CRON:**
image

**POSTPONE 1 DAY and 4AM CRON:**
image

ExtremeFiretop:
This PR now addresses:

1. Router and node with cron offsets
2. Always on-display of the option to download the latest log file
3. Email notification now contains the changelog contents flagged
4. More accurate calculation of expected next run factoring cron and postpone time.

Only thing left would be to add your option to view the MerlinAU logs :)

ExtremeFiretop:
I noticed  a bug with my solution it doesn't properly account for cron days lol.
I was only testing with crons set to run everyday.

Back to the drawing board.

Martinski4GitHub:
> @Martinski4GitHub buddy!
> 
> While I said I wanted to avoid a specific logs menu if possible, I don't know a better way to make it look. My advanced options is getting long, and so is main menu, and I can't put it under mesh nodes for obvious reasons.
> 
> That left me with creating a new menu, but I guess you can use this new menu as a frame for your MerlinAU logs selection option :) Just need a add a new option in the menu for that.
>
OK, sounds good. I expect to finish the menu option to view a selected update log by Saturday evening.

MerlinAU.sh
```diff
@@ -4797,6 +4811,96 @@ _ShowNodesMenu_()
...
+        wget -O "$FW_BIN_DIR/Changelog-${changeLogTag}.txt" "${CL_URL_386}"
+    elif [ "$changeLogTag" = "NG" ]
+    then
+        wget -O "$FW_BIN_DIR/Changelog-${changeLogTag}.txt" "${CL_URL_NG}"
+    fi
+    changeLogFile="${FW_BIN_DIR}/Changelog-${changeLogTag}.txt"
+    if [ ! -f "$changeLogFile" ]
+    then
+        Say "Change-log file [$changeLogFile] does NOT exist."
+        _DoCleanUp_ 1 "$keepZIPfile" "$keepWfile"
```
Martinski4GitHub, line 4848: This same "`_DoCleanUp_`" line is duplicated & executed later in line #4856.

MerlinAU.sh
```diff
@@ -4797,6 +4811,96 @@ _ShowNodesMenu_()
...
+    else
+        clear
+        printf "\n${GRNct}Changelog is ready to review!${NOct}\n"
+        printf "\nPress '${REDct}q${NOct}' to quit when finished.\n"
+        _WaitForEnterKey_
+        less "$changeLogFile"
+    fi
+    _DoCleanUp_ 1 "$keepZIPfile" "$keepWfile"
```
Martinski4GitHub, line 4856: This lines is causing the following error:

`./MerlinAU.sh: line 5019: keepWfile: parameter not set`

Because variable "`keepWfile`" is not defined anywhere.

Also, I don't fully understand the reason to call the"`_DoCleanUp_`" function here just to delete the Changelog file.

ExtremeFiretop: @Martinski4GitHub 

Fixed, I was tested with the Gnuton version :) keepWfile exists there.

MerlinAU.sh
ExtremeFiretop: Removed as requested

ExtremeFiretop:
@Martinski4GitHub 

Fixed

MerlinAU.sh
```diff
@@ -4797,6 +4811,95 @@ _ShowNodesMenu_()
...
+    changeLogFile="${FW_BIN_DIR}/Changelog-${changeLogTag}.txt"
+    if [ ! -f "$changeLogFile" ]
+    then
+        Say "Change-log file [$changeLogFile] does NOT exist."
+    else
+        clear
+        printf "\n${GRNct}Changelog is ready to review!${NOct}\n"
+        printf "\nPress '${REDct}q${NOct}' to quit when finished.\n"
+        _WaitForEnterKey_
+        less "$changeLogFile"
```
Martinski4GitHub, line 4853: I recommend using the "`more`" command instead of `less` to avoid showing the <CR-LF> end of lines:
`more "$changeLogFile"`

This is using `less` cmd:
ChangeLog_ViewWithLess

This is using `more` cmd:
ChangeLog_ViewWithMore

ExtremeFiretop: Good suggestion, I can do that. :)

ExtremeFiretop: @Martinski4GitHub 

Fixed.

Martinski4GitHub (approved):
Good to go.

PR #210, opened 2024-05-19
Code improvement

Martinski4GitHub:
Minor code improvement.


ExtremeFiretop:
I like this.

The functionality is the same, but this offers a more streamlined and cleaner approach to handling the incrementation logic and the construction of the dow_expr variable when the days wrap around the week. This makes it easier to understand even though the core functionality and performance implications between the two are likely the same.

This brings me back to our last discussion about simplicity for the day me and you are no longer working on this, whatever we can do to make it easier to read I'm all for it!

PR #186, opened 2024-04-07
Gnuton Support

ExtremeFiretop:
Gnuton Support (Proof of Concept)

PR #231, opened 2024-06-04
A bug fix and some code improvements.

Martinski4GitHub:
The bug fix in this PR is found on line **4428**.

Basically, in the previous code with only the one conditional check in the "**if**" statement:

`if [ "$fwInstalledBaseVers" -le 3004 ]`

The corresponding "**elif**" conditional section would never be executed because the "**if**" conditional would always take precedence so, in essence, only the "**if**" and "**else**" conditionals would have been executed.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Jun-03
+# Last Modified: 2024-Jun-04
 ###################################################################
 set -u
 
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
I actually thought about this last night while I was sleeping.
Just needed some zzzz's but realized that because I was using -le if the evaluation of fwInstalledBaseVers was true it would never "fail" to go to the elif.

Luckily it matters not, because Martinski already thought of it and fixed it! Thanks buddy!

ExtremeFiretop:
Truth is a did a similar ouchy on the Gnuton branch with this line:

```
           bt) if echo "$PRODUCT_ID" | grep -q "^TUF-"
               then _ChangeBuildType_Gnuton_
               elif [ "$fwInstalledBaseVers" -le 3004 ] || echo "$PRODUCT_ID" | grep -q "^GT-"
               then _ChangeBuildType_Merlin_
               else _InvalidMenuSelection_
               fi
               ;;
```

"OR" is not the logical operator I meant to use, instead I meant to use &&

ExtremeFiretop:
Gnuton synced up here:

https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/186/commits/40acb7f330834d211cb6dfeba9ad38c0a19d472e

PR #232, opened 2024-06-06
Code improvements.

Martinski4GitHub:
- Made code changes/improvements to address an intermittent odd issue when checking for script updates.
- Various other code improvements.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Jun-04
+# Last Modified: 2024-Jun-05
 ###################################################################
 set -u
 
@@ -33,8 +33,8 @@ readonly CL_URL_3006="${FW_URL_BASE}/Documentation/Changelog-3006.txt/download"
 readonly high_risk_terms="factory default reset|features are disabled|break backward compatibility|must be manually|strongly recommended"
 
 # For new script version updates from source repository #
-UpdateNotify=0
 DLRepoVersion=""
+scriptUpdateNotify=0
 
 # For supported version and model checks #
 MinFirmwareCheckFailed=0
@@ -54,10 +54,10 @@ readonly SETTINGSFILE="${SETTINGS_DIR}/custom_settings.txt"
 readonly SCRIPTVERPATH="${SETTINGS_DIR}/version.txt"
 
 ##----------------------------------------##
-## Modified by Martinski W. [2024-May-01] ##
+## Modified by Martinski W. [2024-Jun-05] ##
 ##----------------------------------------##
 ScriptsDirPath="$SCRIPTS_PATH"
-ScriptFilePath="${SCRIPTS_PATH}/$ScriptFileName"
+ScriptFilePath="${SCRIPTS_PATH}/${SCRIPT_NAME}.sh"
 
 if [ ! -f "$ScriptFilePath" ]
 then
@@ -396,7 +396,7 @@ _GetRouterProductID_()
 ##-------------------------------------##
 _ScriptVersionStrToNum_()
 {
-   if [ $# -eq 0 ] || [ -z "$1" ] ; then echo ; return 1 ; fi
+   if [ $# -eq 0 ] || [ -z "$1" ] ; then echo 0 ; return 1 ; fi
    local verNum  verStr
 
    verStr="$(echo "$1" | awk -F '_' '{print $1}')"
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
Hmmm...

I gotta say, off the top of my head this is an odd one... For sure.
Let me have a coffee this morning and think on it throughout my day. 
You know I work best after I've had some time to wake up.

Sadly it's really not much to go on unfortunately, especially because sometimes and works and other times it doesn't based on the description you provided (using Same code)

So naturally I can understand why your thinking it's the curl, but even then I myself have never seen such an issue and it's never been reported in the forums, the curl already has 3 retry attempts, so while I do see this change as an improvement, it's not much of a logical difference and more of a stab in the dark, if it was failing 3 times before, nothing says it won't fail 4 times now and just take longer to fail.

Github is tuned for this type of request, so I have trouble believing Github is having capacity or bandwidth issues and timing out. I do see you made it fail "loudly" now, instead of silently, so maybe all we need to do is wait for the issue again and it will give us a better idea as to what is causing this.

ExtremeFiretop (approved):

ExtremeFiretop:
@Martinski4GitHub 

I was able to recreate the issue... Ish? With our current production version as found below:

image
image

But only by literally breaking our code and deleting the version.txt file, normally I get this if the curl times-out:

image

(No update detected)

This is even if _CheckForNewScriptUpdates_ is timing out and there is an update available.
The reason is that 

These lines:

`   if [ ! -f "$SCRIPTVERPATH" ] ; then UpdateNotify=0 ; return 1 ; fi`
   
   And
   
```
      if [ "$DLRepoVersionNum" -gt "$ScriptVersionNum" ]
   then
      UpdateNotify="New script update available.
${REDct}v$SCRIPT_VERSION${NOct} --> ${GRNct}v$DLRepoVersion${NOct}"
      Say "$(date +'%b %d %Y %X') $(nvram get lan_hostname) ${ScriptFNameTag}_[$$] - INFO: A new script update (v$DLRepoVersion) is available to download."
   else
      UpdateNotify=0
   fi
```
   
Reset the UpdateNotify value to "zero" and the txt file wouldn't be updated as expected.
   
Now for fun I gave it a valid website but an invalid sub-address such as github.com/version1.txt
And got this instead:
   
image

Which is not the issue described by your friend, to me by the symptoms it sounds like the first set of screenshots.
But I don't understand how that is possible since I had to literally delete those checks above I mentioned that reset the UpdateNotify for it to happen.

Next I added a blank version.txt and got this instead:

image

This makes sense since it can't decode a number in the blank file. But I still don't think that's what happened to your friend.

Next I updated the version.txt manually to include a newer version than the script but kept the time-out/block in place, and got this:

image
image

So to me this issue seems to be somehow related to the **UpdateNotify** variable not being reset correctly when the curl times out, or somehow the cat to the $SCRIPTVERPATH variable isn't correctly grabbing the version from the text file..

MerlinAU.sh
```diff
@@ -522,18 +522,22 @@ logo() {
...
    # Download the latest version file from the source repository
-   curl --silent --retry 3 "${SCRIPT_URL_BASE}/version.txt" -o "$SCRIPTVERPATH"
+   curl -LSs --retry 4 --retry-delay 5 "${SCRIPT_URL_BASE}/version.txt" -o "$SCRIPTVERPATH"
 
-   if [ ! -f "$SCRIPTVERPATH" ] ; then UpdateNotify=0 ; return 1 ; fi
+   if [ $? -ne 0 ] || [ ! -s "$SCRIPTVERPATH" ]
+   then scriptUpdateNotify=0 ; return 1 ; fi
 
    # Read in its contents for the current version file
    DLRepoVersion="$(cat "$SCRIPTVERPATH")"
```
ExtremeFiretop, line 543: Any thoughts on added a check for this variable?

    # Read the contents of the version file
    DLRepoVersion="$(cat "$SCRIPTVERPATH" 2>/dev/null)"
    if [ -z "$DLRepoVersion" ]; then
        echo "Downloaded version file is empty."
        UpdateNotify=0
        return 1
    fi

ExtremeFiretop:
@Martinski4GitHub 

Based on my gut feeling telling me somehow it's related to the `DLRepoVersion="$(cat "$SCRIPTVERPATH")"`
My suggestion was to add a check on that variable.

Thoughts?

Martinski4GitHub:
> Hmmm...
> 
> I gotta say, off the top of my head this is an odd one... For sure. Let me have a coffee this morning and think on it throughout my day. You know I work best after I've had some time to wake up, usually I work best when stuff is let sitting in the background of my brain for a while.
> 
> Sadly it's really not much to go on unfortunately, especially because sometimes it works and other times it doesn't based on the description you provided (using Same code)
> 
> So naturally I can understand why your thinking it's the curl, but even then I myself have never seen such an issue and it's never been reported in the forums, the curl already has 3 retry attempts, so while I do see this change as an improvement, it's not much of a logical difference and more of a stab in the dark, if it was failing 3 times before, nothing says it won't fail 4 times now and just take longer to fail.
> 
> Github is tuned for this type of request, so I have trouble believing Github is having capacity or bandwidth issues and timing out, and then all of a sudden on a retry it works? I do see you made it fail "loudly" now, instead of silently, so maybe all we need to do is wait for the issue again and it will give us a better idea as to what is causing this.
>
Yes, I agree. It was odd and a very unlikely scenario that the `curl` command fails (or as I granted before: "very rare"), but it's not completely impossible. Also, note that the improvements were more than just changing the `curl` command options/switches to get one more retry, have a longer "retry-delay" to 5 seconds, and show errors.

For example, the previous check code was like this:

`if [ ! -f "$SCRIPTVERPATH" ] ; then UpdateNotify=0 ; return 1 ; fi`

Which checks only if the "version.txt" file exists.

The new improved code is like this:

```
if [ $? -ne 0 ] || [ ! -s "$SCRIPTVERPATH" ]
then scriptUpdateNotify=0 ; return 1 ; fi
```

This is a significant difference in the logic because it checks for 3 things:

1) The return code of the `curl` command.
2) If the "version.txt" file exists.
3) If the "version.txt" file is empty.

In any event, I'm hoping that by adding those extra options, more conditional checks, and reporting any curl errors, we are handling things a bit better, and perhaps if it happens again, the reported error may provide more clues as to the cause is.

Martinski4GitHub:
> 
> So to me this issue seems to be somehow related to the **UpdateNotify** variable not being reset correctly when the curl times out, but more likely, is somehow related to the cat to the $SCRIPTVERPATH variable isn't correctly populating the **DLRepoVersion** variable.
>
Yes, that's another possibility.

MerlinAU.sh
Martinski4GitHub: > Any thoughts on added a check for this variable?
> 
> ```
> # Read the contents of the version file
> DLRepoVersion="$(cat "$SCRIPTVERPATH" 2>/dev/null)"
> if [ -z "$DLRepoVersion" ]; then
>     echo "Downloaded version file is empty."
>     UpdateNotify=0
>     return 1
> fi
> ```
>
I don't have any objections to adding the extra check. However, note that the newly-merged code already checks if the "version.txt" file is empty. So this extra check would be if the `cat` command fails somehow, so the error message is not that the file is empty, but that the variable is empty.

`echo "Variable for downloaded version is empty."`

My 2 cents.

ExtremeFiretop:
Synced with Gnuton here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/186/commits/535dbcd41fcbea30f262521555b239f37a010da5

PR #233, opened 2024-06-07
Update Curl to WGET

ExtremeFiretop:
wget's major strong side compared to curl is its ability to download recursively. (Curl cannot)
wget is command line only. There's no lib or anything, but curl's features are powered by libcurl.
curl offers upload and sending capabilities which is the main difference between curl and wget...
wget is only used to download things, while curl is used to test communication by using a specific protocol, etc.

This actually lines up with what we use in the Gnuton branch to interact with Github if you hadn't already noticed.
The standard in the script I was going for seems to be to use curl when we need to simulate a browser and need advanced "browser" features to "test" things. But that's up to discussion.

And wget is used just to grab/download files from Github. Thoughts?
This discussion will also determine what I do about all the wgets in the Gnuton branch, which does not exist in the standard branch (all curls in production).

I recommend you take a quick look that the solution there for grabbing the TUF UIs from Github.
Let me know your 2 cent's. I'm looking forwards :)

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -534,13 +534,18 @@ _CheckForNewScriptUpdates_()
    rm -f "$SCRIPTVERPATH"
 
    # Download the latest version file from the source repository
-   curl -LSs --retry 4 --retry-delay 5 "${SCRIPT_URL_BASE}/version.txt" -o "$SCRIPTVERPATH"
+   wget --quiet --tries=4 --waitretry=5 "${SCRIPT_URL_BASE}/version.txt" -O "$SCRIPTVERPATH"
 
    if [ $? -ne 0 ] || [ ! -s "$SCRIPTVERPATH" ]
    then scriptUpdateNotify=0 ; return 1 ; fi
 
    # Read in its contents for the current version file
    DLRepoVersion="$(cat "$SCRIPTVERPATH")"
+   if [ -z "$DLRepoVersion" ]; then
+       echo "Variable for downloaded version is empty."
+       UpdateNotify=0
+       return 1
+   fi
 
    DLRepoVersionNum="$(_ScriptVersionStrToNum_ "$DLRepoVersion")"
    ScriptVersionNum="$(_ScriptVersionStrToNum_ "$SCRIPT_VERSION")"
@@ -582,8 +587,8 @@ _SCRIPTUPDATE_()
       then
           echo ; echo
           echo -e "${CYANct}Downloading $SCRIPT_NAME ${CYANct}v$DLRepoVersion${NOct}"
-          curl -LSs --retry 4 --retry-delay 5 "${SCRIPT_URL_BASE}/version.txt" -o "$SCRIPTVERPATH"
-          curl -LSs --retry 4 --retry-delay 5 "${SCRIPT_URL_BASE}/${SCRIPT_NAME}.sh" -o "$ScriptFileDL"
+          wget --quiet --tries=4 --waitretry=5 "${SCRIPT_URL_BASE}/version.txt" -O "$SCRIPTVERPATH"
+          wget --quiet --tries=4 --waitretry=5 "${SCRIPT_URL_BASE}/${SCRIPT_NAME}.sh" -O "$ScriptFileDL"
 
           if [ $? -eq 0 ] && [ -s "$ScriptFileDL" ]
           then
@@ -613,8 +618,8 @@ _SCRIPTUPDATE_()
       then
           echo ; echo
           echo -e "${CYANct}Downloading $SCRIPT_NAME ${CYANct}v$DLRepoVersion${NOct}"
-          curl -LSs --retry 4 --retry-delay 5 "${SCRIPT_URL_BASE}/version.txt" -o "$SCRIPTVERPATH"
-          curl -LSs --retry 4 --retry-delay 5 "${SCRIPT_URL_BASE}/${SCRIPT_NAME}.sh" -o "$ScriptFileDL"
+          wget --quiet --tries=4 --waitretry=5 "${SCRIPT_URL_BASE}/version.txt" -O "$SCRIPTVERPATH"
+          wget --quiet --tries=4 --waitretry=5 "${SCRIPT_URL_BASE}/${SCRIPT_NAME}.sh" -O "$ScriptFileDL"
 
           if [ $? -eq 0 ] && [ -s "$ScriptFileDL" ]
           then
@@ -1594,7 +1599,7 @@ _SendEMailNotification_()
 
    date +"$LOGdateFormat" > "$userTraceFile"
 
-   /usr/sbin/curl -Lv --retry 4 --retry-delay 5 --url "${PROTOCOL}://${SMTP}:${PORT}" \
+   curl -Lv --retry 4 --retry-delay 5 --url "${PROTOCOL}://${SMTP}:${PORT}" \
    --mail-from "$FROM_ADDRESS" --mail-rcpt "$TO_ADDRESS" $CC_ADDRESS_ARG \
    --user "${USERNAME}:$(/usr/sbin/openssl aes-256-cbc "$emailPwEnc" -d -in "$amtmMailPswdFile" -pass pass:ditbabot,isoi)" \
    --upload-file "$tempEMailContent" \


ExtremeFiretop:
And btw this is just a suggestion.

I want to know your stance on if we should be using the same standard throughout the script (curl for everything) or we should default to the simpler wget unless we need the curl?

Thoughts, open to discussion?

Martinski4GitHub:
> 
> wget's major strong side compared to curl is its ability to download recursively. (Curl cannot) 
>
Yes, that's true. But how does that really help us when downloading one small file at a time?
(version.txt is less than 1KB. The shell script itself is ~210KB)

>
> wget is command line only. There's no lib or anything, but curl's features are powered by libcurl. 
>
OK...?... I'm not sure what you're trying to say here (`curl` is a CLI tool as well), and I don't know what that implies in terms of actual functionality. Are you saying that `wget` is much more robust and less likely to fail than the `curl` command?

>
> curl offers upload and sending capabilities which is the main difference between curl and wget... wget is only used to download things, while curl is used to test communication by using a specific protocol, etc.
> 
> This actually lines up with what we use in the Gnuton branch to interact with Github if you hadn't already noticed. The standard in the script I was going for seems to be to use curl when we need to simulate a browser and need advanced "browser" features to "test" things. But that's up to discussion.
> 
> And wget is used just to grab/download files from Github. Thoughts? This discussion will also determine what I do about all the wgets in the Gnuton branch, which does not exist in the standard branch (all curls in production).
> 
> I recommend you take a quick look that the solution there for grabbing the TUF UIs from Github. Let me know your 2 cent's, I'm waiting for it.
>
>
Honestly, I have no real preference for either command. Certainly, to upload/send files `curl` is the method to use; but for downloading single files from URLs, I have used both `curl` & `wget` and have found no discernible advantage in using one over the other.

WRT the "**quiet**" option:

`wget --quiet ...`

I'd rather not go back to "silent" failures again. In the unlikely event that the download fails, I still like to see whatever error message is reported so that, at least, we have a clue as to what the problem may be during interactive sessions.

My 2 cents.

MerlinAU.sh
```diff
@@ -534,13 +534,18 @@ _CheckForNewScriptUpdates_()
...
+   wget --quiet --tries=4 --waitretry=5 "${SCRIPT_URL_BASE}/version.txt" -O "$SCRIPTVERPATH"
 
    if [ $? -ne 0 ] || [ ! -s "$SCRIPTVERPATH" ]
    then scriptUpdateNotify=0 ; return 1 ; fi
 
    # Read in its contents for the current version file
    DLRepoVersion="$(cat "$SCRIPTVERPATH")"
+   if [ -z "$DLRepoVersion" ]; then
+       echo "Variable for downloaded version is empty."
+       UpdateNotify=0
```
Martinski4GitHub, line 546: BTW, the variable name was changed to "`scriptUpdateNotify`" to make it distinct & clear that it's not related to the F/W Update notification.

ExtremeFiretop:
> Honestly, I have no real preference for either command. Certainly, to upload/send files `curl` is the method to use; but for downloading single files from URLs, I have used both `curl` & `wget` and have found no discernible advantage in using one over the other.
> 

I think the general idea was to get a consensus for a standard moving forward as mentioned here:

> And btw this is just a suggestion.
> 
> I want to know your stance on if we should be using the same standard throughout the script (curl for everything) or we should default to the simpler wget unless we need the curl?
> 
> Thoughts, open to discussion?

Since my method diverged with Gnuton compared to production.
Unfortunately it seems you have no preference so I guess use whatever you rather use at the moment I guess :P

> WRT the "**quiet**" option:
> 
> `wget --quiet ...`
> 
> I'd rather not go back to "silent" failures again. In the unlikely event that the download fails, I still like to see whatever error message is reported so that, at least, we have a clue as to what the problem may be during interactive sessions.
> 
> My 2 cents.

Just as an FYI, wget --quiet will suppress most of the output, including the progress bar and normal messages, but it will still display error messages.

Martinski4GitHub:
> > Honestly, I have no real preference for either command. Certainly, to upload/send files `curl` is the method to use; but for downloading single files from URLs, I have used both `curl` & `wget` and have found no discernible advantage in using one over the other.
> 
> I think the general idea was to get a consensus for a standard moving forward as mentioned here:
> 
> > And btw this is just a suggestion.
> > I want to know your stance on if we should be using the same standard throughout the script (curl for everything) or we should default to the simpler wget unless we need the curl?
> > Thoughts, open to discussion?
> 
Ah, OK. It was not crystal clear that's what you were going for. It seems your initial message was trying to promote `wget` as a better alternative than `curl` but I saw no evidence that it was technically better than the other (recursive downloading?? wget as command line only - no lib??).

In any case, I have no objections to making `curl` or `wget` the "standard" command to call based on the required usage as long we don't lose the ability to get the reported errors during interactive sessions.

>
> > And wget is used just to grab/download files from Github. Thoughts? This discussion will also determine what I do about all the wgets in the Gnuton branch, which does not exist in the standard branch (all curls in production).
> 
> Since my method diverged with Gnuton compared to production. Unfortunately it seems you have no preference so I guess use whatever you rather use at the moment I guess :P
> 
> > WRT the "**quiet**" option:
> > `wget --quiet ...`
> > I'd rather not go back to "silent" failures again. In the unlikely event that the download fails, I still like to see whatever error message is reported so that, at least, we have a clue as to what the problem may be during interactive sessions.
> > My 2 cents.
> 
> Just as an FYI, wget --quiet will suppress most of the output, including the progress bar and normal messages, but it will still display error messages.
>
OK, this is good to know and removes my objection to the "**quiet**" option.

ExtremeFiretop:
> > wget's major strong side compared to curl is its ability to download recursively. (Curl cannot)
> 
> Yes, that's true. But how does that really help us when downloading one small file at a time? (version.txt is less than 1KB. The shell script itself is ~210KB)
> 

I think your viewing this request too much as a change for this specific file, instead of the goal which was to get a consensus/standard for any "download". No matter the file. 
In future instances, we may want the recursive feature, or we may want the simplicity of the wget implementation. Etc.

> > wget is command line only. There's no lib or anything, but curl's features are powered by libcurl.
> 
> OK...?... I'm not sure what you're trying to say here (`curl` is a CLI tool as well), and I don't know what that implies in terms of actual functionality. Are you saying that `wget` is much more robust and less likely to fail than the `curl` command?

it was mostly about comparing both tools, and determining which should be the "default" in the script so I can update either production or Gnuton to match this standard since it diverged.

Should we be using curls everywhere since it's more advanced and can do what wget cant? Or should we use wget everywhere and only use curl when wget is limited.

But yes my understanding is that curl due to being a tool with lib, is more subject to "changes" than wget is. Not even talking about "bugs" in curl, but just changes in parameters, etc. while wget is more "integrated"? and therefore we can "expect" will maintain the same undisturbed functionality longer.

This was to be an open discussion on the theories around both but it seems there's no preference so I'll just continue to use whatever at the time, and won't feel inclined to update Gnuton or production to some type of sta dard to match.

Hope you have a good weekend bud, chat soon. 🙂

ExtremeFiretop:
> > > Honestly, I have no real preference for either command. Certainly, to upload/send files `curl` is the method to use; but for downloading single files from URLs, I have used both `curl` & `wget` and have found no discernible advantage in using one over the other.
> > 
> > 
> > I think the general idea was to get a consensus for a standard moving forward as mentioned here:
> > > And btw this is just a suggestion.
> > > I want to know your stance on if we should be using the same standard throughout the script (curl for everything) or we should default to the simpler wget unless we need the curl?
> > > Thoughts, open to discussion?
> 
> Ah, OK. It was not crystal clear that's what you were going for. It seems your initial message was trying to promote `wget` as a better alternative than `curl` but I saw no evidence that it was technically better than the other (recursive downloading?? wget as command line only - no lib??).
> 
> In any case, I have no objections to making `curl` or `wget` the "standard" command to call based on the required usage as long we don't lose the ability to get the reported errors during interactive sessions.
> 
> > > And wget is used just to grab/download files from Github. Thoughts? This discussion will also determine what I do about all the wgets in the Gnuton branch, which does not exist in the standard branch (all curls in production).
> > 
> > 
> > Since my method diverged with Gnuton compared to production. Unfortunately it seems you have no preference so I guess use whatever you rather use at the moment I guess :P
> > > WRT the "**quiet**" option:
> > > `wget --quiet ...`
> > > I'd rather not go back to "silent" failures again. In the unlikely event that the download fails, I still like to see whatever error message is reported so that, at least, we have a clue as to what the problem may be during interactive sessions.
> > > My 2 cents.
> > 
> > 
> > Just as an FYI, wget --quiet will suppress most of the output, including the progress bar and normal messages, but it will still display error messages.
> 
> OK, this is good to know and removes my objection to the "**quiet**" option.

I just saw this now, but I think my latest reply answers this though. I hope my latest reply with more clear on the goal of the discussion.

Martinski4GitHub:
> 
> I just saw this now, but I think my latest reply answers this though. I hope my latest reply with more clear on the goal of the discussion.
>
>
Yeah, your next-to-last message was much clearer WRT your intended goal. Sometimes, we just need to elaborate a bit more to convey the full meaning.

Enjoy your weekend, bud.

PR #236, opened 2024-06-10
Fix/Patch Cron Estimates

ExtremeFiretop:
Fix/Patch Cron Estimates

Currently this cron: 0 0 * * * (our default)
does not work with our production estimates.

The below fixes this. Seems the issue lies in the logic for checking the hour and minute fields.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,11 +4,11 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Jun-05
+# Last Modified: 2024-Jun-10
 ###################################################################
 set -u
 
-readonly SCRIPT_VERSION=1.2.1
+readonly SCRIPT_VERSION=1.2.2
 readonly SCRIPT_NAME="MerlinAU"
 
 ##-------------------------------------##
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -3106,9 +3106,9 @@ expand_cron_field()
     fi
 }
 
-##----------------------------------------##
-## Modified by Martinski W. [2024-May-18] ##
-##----------------------------------------##
+##------------------------------------------##
+## Modified by ExtremeFiretop [2024-Jun-10] ##
+##------------------------------------------##
 _EstimateNextCronTimeAfterDate_()
 {
     local post_date_secs="$1"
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
@Martinski4GitHub 

I think this should be an emergency release.
Not good that the default cron can't be calculated, luckily there's been no new updates so it's gone unreported yet.

MerlinAU.sh
```diff
@@ -3139,17 +3139,18 @@ _EstimateNextCronTimeAfterDate_()
...
         then
             for this_hour in $(expand_cron_field "$hour_field" 0 23)
             do
-                if [ "$this_hour" -gt "$current_hour" ]
+                if [ "$this_hour" -gt "$current_hour" ] || [ "$loopCount" -gt 1 ]
                 then
                     for this_min in $(expand_cron_field "$minute_field" 0 59)
                     do
-                        echo "$(date '+%s' -d "$current_year-$current_month-$current_day $this_hour:$this_min")"
+                        echo "$(date -d "@$(date '+%s' -d "$current_year-$current_month-$current_day $this_hour:$this_min")" '+%Y-%m-%d %H:%M:%S')"
```
Martinski4GitHub, line 3153: It looks like you changed the format of the value being reported & eventually how it's stored in the configuration file: from a raw number of seconds to human-readable date/time stamp. I like it!!

ExtremeFiretop: I did it for 2 reasons.

It seemed redundant to store in seconds and converted back. (It was less code to do it this way)

And second it follows the other date formats in the configuration file if you happen it open it up.

Martinski4GitHub:
> @Martinski4GitHub
> 
> I think this should be an emergency release. Not good that the default cron can't be calculated, luckily there's been no new updates so it's gone unreported yet.
>
Agreed. I'm currently reviewing & testing the fix to validate. I'm still not sure how the fix works, but it does :>).

ExtremeFiretop:
> > @Martinski4GitHub
> > I think this should be an emergency release. Not good that the default cron can't be calculated, luckily there's been no new updates so it's gone unreported yet.
> 
> Agreed. I'm currently reviewing & testing the fix to validate. I'm still not sure how the fix works, but it does :>).

image

MerlinAU.sh
Martinski4GitHub: >
> 
> And second it follows the other date formats in the configuration file if you happen it open it up.
>
Yeah, the "F/W Update Notification Date" is stored in human-readable form as well so they're consistent.

MerlinAU.sh
```diff
@@ -3139,17 +3139,18 @@ _EstimateNextCronTimeAfterDate_()
...
         loopCount="$((loopCount + 1))"
+
         if matches_month "$current_month" "$month_field" && \
            matches_day_of_month "$current_day" "$dom_field" && \
            matches_day_of_week "$current_dow" "$dow_field"
         then
             for this_hour in $(expand_cron_field "$hour_field" 0 23)
             do
-                if [ "$this_hour" -gt "$current_hour" ]
+                if [ "$this_hour" -gt "$current_hour" ] || [ "$loopCount" -gt 1 ]
```
Martinski4GitHub, line 3149: Ah, OK. The check for "`$loopCount" -gt 1`" essentially bypasses the rest of the code, which is why it works.

Martinski4GitHub (approved):

ExtremeFiretop:
@Martinski4GitHub 

Let me know if I should ready a release if you want to do anything else before a release.

MerlinAU.sh
Martinski4GitHub: > Ah, OK. The check for "`$loopCount" -gt 1`" essentially bypasses the rest of the code, which is why it works.
>
The more I think about it, the more I come to realize that bypassing the rest of the code when the `loopCount` reaches 2 
is not quite the right fix. I think the fix is further down to checking the minutes.

I'll submit a PR with my proposed solution and see how it works for you.

ExtremeFiretop: @Martinski4GitHub 

Received.

I'll let the master cook, I'll check back later! 😉

PR #241, opened 2024-06-21
ROG Patch for 388.8 and Changelog Notifications

ExtremeFiretop:
ROG Patch for 388.8 and Changelog Notifications

As discussed in the Gnuton branch, I've noticed some things while testing the latest alpha.
The 388 alpha, It no longer contains ROG:
image

So while RMerlin said hes dropping support in 3006, I guess he meant just going forwards no more support.
https://www.snbforums.com/threads/3004-388-8-alpha-build-s-testing-available-build-s.90624/post-914160
Apparently, since 3004 and 3006 use the same build script.

And this doesn't look right to me:
image

This is my quick quick assumption of what needs to be done, but I've tested nothing, I'll be back Monday to check in :)
Have a great weekend Martinski!

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -1173,7 +1173,8 @@ _Init_Custom_Settings_Config_
 # ROG upgrades to 3006 codebase should have 
 # the ROG option deleted.
 #-----------------------------------------------------------
-if [ "$fwInstalledBaseVers" -ge 3006 ] && grep -q "^ROGBuild" "$SETTINGSFILE"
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
The changelog notifications only happened because I ran merlinAU with no update detected, and then flipped the switch on: nvram set webs_state_flag=1 while it was running and then hit option 1, this would never happen with a regular user, it would detect the update when first loading before getting to the main menu, or as the cron job before hand, but I just figured I'd see if I could limit this unintended behaviour.

ExtremeFiretop:
ROG options are back: https://www.snbforums.com/threads/asuswrt-merlin-3004-388-8-beta-is-now-available.90713/

Will re-evaluate when I'm back home

PR #242, opened 2024-06-24
Fix Changelog Check Behavior

ExtremeFiretop:
Fix Changelog Check Behavior.

As identified in the Gnuton branch, this shouldn't be happening, this PR/change continues to allow everything to function correctly when the cron detects an update or when an interactive session detects an update. 

But will limit it from running when an update was detected while MerlinAU was already running at the interactive screen.
While this is unlikely to happen to a user, since I can only make it happen when manually running: `nvram set webs_state_flag=1 ` it will make the debugging scenarios run as expected and limit unintended behavior.

image

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Jun-16
+# Last Modified: 2024-Jun-24
 ###################################################################
 set -u
 
@@ -93,6 +93,7 @@ fi
 ##----------------------------------------##
 inMenuMode=true
 isInteractive=false
+FlashStarted=false
 
 readonly mainLAN_IPaddr="$(nvram get lan_ipaddr)"
 readonly fwInstalledBaseVers="$(nvram get firmver | sed 's/\.//g')"
@@ -1173,7 +1174,8 @@ _Init_Custom_Settings_Config_
 # ROG upgrades to 3006 codebase should have 
 # the ROG option deleted.
 #-----------------------------------------------------------
-if [ "$fwInstalledBaseVers" -ge 3006 ] && grep -q "^ROGBuild" "$SETTINGSFILE"
+if { [ "$fwInstalledBaseVers" -ge 3006 ] && grep -q "^ROGBuild" "$SETTINGSFILE"; } || 
+   { [ "$fwInstalledBaseVers" -eq 3004 ] && [ "$(echo "$fwInstalledBaseVers 388.8" | awk '{print ($1 > $2)}')" -eq 1 ] && grep -q "^ROGBuild" "$SETTINGSFILE"; }
 then
     Delete_Custom_Settings "ROGBuild"
 fi
@@ -1753,7 +1755,7 @@ _GetCurrentFWInstalledLongVersion_()
 {
 
 ##FOR TESTING/DEBUG ONLY##
-if false ; then echo "3004.388.6.2" ; return 0 ; fi
+if true ; then echo "3004.388.6.2" ; return 0 ; fi
 ##FOR TESTING/DEBUG ONLY##
 
    local theVersionStr  extVersNum
@@ -3815,7 +3817,10 @@ _CheckNewUpdateFirmwareNotification_()
            then
               _SendEMailNotification_ NEW_FW_UPDATE_STATUS
            fi
-           _ManageChangelog_ "download" "$fwNewUpdateNotificationVers"
+           if ! "$FlashStarted"
+           then
+              _ManageChangelog_ "download" "$fwNewUpdateNotificationVers"
+           fi
        fi
    fi
 
@@ -3828,7 +3833,10 @@ _CheckNewUpdateFirmwareNotification_()
        then
           _SendEMailNotification_ NEW_FW_UPDATE_STATUS
        fi
-       _ManageChangelog_ "download" "$fwNewUpdateNotificationVers"
+       if ! "$FlashStarted"
+       then
+          _ManageChangelog_ "download" "$fwNewUpdateNotificationVers"
+       fi
    fi
 
    fwNewUpdateNotificationDate="$(Get_Custom_Setting FW_New_Update_Notification_Date)"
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
>
> Fix Changelog Check Behavior.
> 
> As identified in the Gnuton branch, this shouldn't be happening, this PR/change continues to allow everything to function correctly when the cron detects an update or when an interactive session detects an update.
> 
> But will limit it from running when an update was detected while MerlinAU was already running at the interactive screen. While this is unlikely to happen to a user, since I can only make it happen when manually running: `nvram set webs_state_flag=1 ` it will make the debugging scenarios run as expected and limit unintended behavior.
> 
>
Honestly, I'm not sure I fully understand the use case you described. I believe what you're saying is that when the user selects the "**Run F/W Update Check Now**" option from the Main Menu, the "**F/W Changelog Check**" should not occur under certain conditions, even if it's **ENABLED**?

Is my understanding correct? If so, what are those specific conditions? 
I'm still trying to fully grasp the scenario you're trying to address.

ExtremeFiretop:
> > Fix Changelog Check Behavior.
> > As identified in the Gnuton branch, this shouldn't be happening, this PR/change continues to allow everything to function correctly when the cron detects an update or when an interactive session detects an update.
> > But will limit it from running when an update was detected while MerlinAU was already running at the interactive screen. While this is unlikely to happen to a user, since I can only make it happen when manually running: `nvram set webs_state_flag=1 ` it will make the debugging scenarios run as expected and limit unintended behavior.
> 
> Honestly, I'm not sure I fully understand the use case you described. I believe what you're saying is that when the user selects the "**Run F/W Update Check Now**" option from the Main Menu, the "**F/W Changelog Check**" should not occur under certain conditions, even if it's **ENABLED**?
> 
> Is my understanding correct? If so, what are those specific conditions? I'm still trying to fully grasp the scenario you're trying to address.

The prompt should not show there, that is only to show when starting the script. Not when running the update :)
You can tell by the "press enter to return to main menu" which when you hit enter, continues the update.

ExtremeFiretop:
One second I'll grab some more screenshots to show a better comparison.

ExtremeFiretop:
@Martinski4GitHub 

This is normal behavior when that prompt should show:
image

And hitting enter should bring you to the main menu:
image

And as you can see here below, even though it's enabled and "blocked" it does not show the prompt when running the update:
image

Instead it shows it here:
image

What you see in the original screenshot is not normal behavior for that prompt to show up. I am limiting this behavior, which is possible if MerlinAU was already running at the interactive menu, when the switch to: `nvram set webs_state_flag=1` is flipped.

Not standard for a normal user, that switch would flip when first running MerlinAU interactively or as a cron normally. Not while sitting at the interactive menu like I did, but I am still trying to limit unintended behavior

Martinski4GitHub:
> @Martinski4GitHub
> 
> This is normal behavior when that prompt should show: **(Right when starting MerlinAU)** 
> And hitting enter should bring you to the main menu: 
> ...
> 
> And as you can see here below, even though the changelog check is enabled and "blocked", it does not show the prompt when running the update: 
> ...
> 
> Instead it shows it here: 
> ...
> 
> What you see in the original screenshot is not normal behavior for that prompt to show up. I am limiting this behavior, which is possible if MerlinAU was already running at the interactive menu, when the switch to: `nvram set webs_state_flag=1` is flipped.
> 
> Not standard for a normal user, that switch would flip when first running MerlinAU interactively or as a cron normally. Not while sitting at the interactive menu like I did, but I am still trying to limit unintended behavior
>
Ah, OK. I think I really got it now. The additional screenshots are very useful to see the "whole picture."
Thanks for the clarification. Sounds like a good idea to "limit unintended behavior" as you said.

Martinski4GitHub (approved):

ExtremeFiretop:
> > @Martinski4GitHub
> > This is normal behavior when that prompt should show: **(Right when starting MerlinAU)**
> > And hitting enter should bring you to the main menu:
> > ...
> > And as you can see here below, even though the changelog check is enabled and "blocked", it does not show the prompt when running the update:
> > ...
> > Instead it shows it here:
> > ...
> > What you see in the original screenshot is not normal behavior for that prompt to show up. I am limiting this behavior, which is possible if MerlinAU was already running at the interactive menu, when the switch to: `nvram set webs_state_flag=1` is flipped.
> > Not standard for a normal user, that switch would flip when first running MerlinAU interactively or as a cron normally. Not while sitting at the interactive menu like I did, but I am still trying to limit unintended behavior
> 
> Ah, OK. I think I really got it now. The additional screenshots are very useful to see the "whole picture." Thanks for the clarification. Sounds like a good idea to "limit unintended behavior" as you said.

No worries I'm happy I could help clarify buddy 😃

The only way I could see this happening to a user is real life, is if a user happened to go to bed with MerlinAU running in their SSH terminal still. And the routers automated run of the webs_update.sh script flips the switch could it found an update available while the user was sleeping with the script "forgot" running for extended amount of time.

But again, even if it's not normal or likely to be caught I just figured I'd patch it while I ran into it.

Martinski4GitHub:
> > > @Martinski4GitHub
> > > This is normal behavior when that prompt should show: **(Right when starting MerlinAU)**
> > > And hitting enter should bring you to the main menu:
> > > ...
> > > And as you can see here below, even though the changelog check is enabled and "blocked", it does not show the prompt when running the update:
> > > ...
> > > Instead it shows it here:
> > > ...
> > > What you see in the original screenshot is not normal behavior for that prompt to show up. I am limiting this behavior, which is possible if MerlinAU was already running at the interactive menu, when the switch to: `nvram set webs_state_flag=1` is flipped.
> > > Not standard for a normal user, that switch would flip when first running MerlinAU interactively or as a cron normally. Not while sitting at the interactive menu like I did, but I am still trying to limit unintended behavior
> > 
> > 
> > Ah, OK. I think I really got it now. The additional screenshots are very useful to see the "whole picture." Thanks for the clarification. Sounds like a good idea to "limit unintended behavior" as you said.
> 
> No worries I'm happy I could help clarify buddy 😃
> 
> The only way I could see this happening to a user in real life, is if a user happened to go to bed with MerlinAU running in their SSH terminal still. And the routers automated run of the webs_update.sh script flips the switch could it found an update available while the user was sleeping with the script "forgot" running for extended amount of time.
> 
> But again, even if it's not normal or likely to be caught I just figured I'd patch it while I ran into it.
>
Yep, fully agreed. A preemptive move :>).

PR #249, opened 2024-06-28
Small Improvements

ExtremeFiretop:
Small Improvements

PR #250, opened 2024-06-28
Fixes to PR #249.

Martinski4GitHub:


MerlinAU.sh
```diff
@@ -4142,7 +4143,13 @@ _EntwareServicesHandler_()
        done
        return 0
    }
-   
+
+   case "$1" in
```
ExtremeFiretop, line 4147: This is already fixed :)

MerlinAU.sh
```diff
@@ -4117,6 +4117,7 @@ _EntwareServicesHandler_()
    if [ $# -eq 0 ] || [ -z "$1" ] ; then return 1 ; fi
    local AllowVPN="$(Get_Custom_Setting Allow_Updates_OverVPN)"
 
+   local actionStr=""
```
ExtremeFiretop, line 4120: Good point on adding this though, I don't think I did.

MerlinAU.sh
```diff
@@ -4739,7 +4746,7 @@ Please manually update to version $minimum_supported_version or higher to use th
            then
                # Diversion unmount command also unloads entware services #
                Say "Stopping Diversion service..."
-               /opt/bin/diversion unmount >/dev/null
+               /opt/bin/diversion unmount &
```
ExtremeFiretop, line 4749: Why are we running this in the background again?
Shouldn't we wait for diversion to unmount before we start flashing?

ExtremeFiretop: Oh wait. Nvm I went back through the PRs and it was cause Gnuton was crashing and freezing when trying to unmount.

Yes yes. My memory is coming back to me now hahah

ExtremeFiretop: 2 nights in a row of little sleep is starting to showwww.

Can we still hide the output though? So we don't see the diversion splash right before the flash.

Martinski4GitHub: > Why are we running this in the background again? Shouldn't we wait for diversion to unmount before we start flashing?
>
If Diversion has not completed & returned after 5 seconds, something has gone wrong, at which point we don't want to hang the F/W flash process because of it. We just continue to eject the USB drive and move on. Remember that by this point we've already stopped all other Entware services except for Diversion & Tailscale (if enabled).

Martinski4GitHub: > 2 nights in a row of little sleep is starting to showwww.
> 
> Can we still hide the output though? So we don't see the diversion splash right before the flash.
>
I thought it was good for the user to see that Diversion has been unmounted when running interactively.
But either way is fine with me. I have no preference.

ExtremeFiretop:
@Martinski4GitHub 

Merged, and removed the duplicate here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/commit/bad5daf5381a092480130b25ecff39defff520ec

MerlinAU.sh
ExtremeFiretop: Merged and removed the duplicate here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/commit/bad5daf5381a092480130b25ecff39defff520ec

ExtremeFiretop:
Synced with Gnuton here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/186/commits/783a4a6e65fee3e67df5511aa0b436b0e2c147a6

PR #271, opened 2024-07-24
Minor change to the "Help" output

Martinski4GitHub:
Added the "**.sh**" file extension to the listed script commands in the "Help" output screen to make it clear that the extension is definitely needed as part of the script filename.

**BEFORE**:
MerlinAU_HelpCmds_BEFORE

**AFTER**:
MerlinAU_HelpCmds_AFTER

PR #272, opened 2024-07-24
Update MerlinAU.sh

ExtremeFiretop:


Martinski4GitHub:
@ExtremeFiretop,

I have not yet had a chance to review the latest changes in this PR (I was busy this evening rebuilding the main drive from my wife's laptop & reinstalling all her apps from scratch - it's all good now). But from the SNB Forums post you referenced asking for the "Manual Update" feature, I still have the same concerns you listed in your reply about allowing the user to do a "manual update" using a F/W ZIP file that was previously downloaded locally.

My main concern is that the local ZIP file containing the F/W image can be of **unknown** origin, from an **unknown** source, even a Beta or Alpha build, and this opens a potential "can of worms" where the router can be accidentally bricked by trying to flash an image that may or may not have been "vetted & blessed" by RMerlin. Currently, our script downloads the image directly from RMerlin's website so we are extremely confident that the image is "good" and we are now verifying its authenticity & integrity by checking the SHA256 signature against RMerlin's official firmware signature.

Now, from a developer's point of view, I can see where this "manual update" feature would be useful if you wanted to build your own F/W images and then flash them into the target router. But this is essentially a "backdoor" for development purposes only; I would not want normal, regular users (many of whom lack the necessary technical skills to know better) to have the ability to flash a potentially **unknown** & **unvetted** F/W image file using our script, and if something goes wrong, blame our script for unforeseen consequences.

IMO, if a user really wants to flash a F/W image that can potentially be from an "**unknown**" source, they can always use the built-in webGUI. I understand that the OP from the SNBForum post was using TailScale to connect remotely and his concern was that the connection was terminated ***before*** the F/W flash was fully completed. I think this has been already addressed in our latest script release (BTW, I'll submit a PR that should improve this feature, based on our recent discussions in PR #271).

ExtremeFiretop:
Sorry, I meant to update this PR, not merge it lol.
I've reverted it.

ExtremeFiretop:
Opened it again here to continue the discussion:

https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/275

PR #273, opened 2024-07-25
Allow script to continue flashing F/W if connection is terminated.

Martinski4GitHub:
Made changes to allow script to continue with the F/W flash "`curl`" command after connection is terminated due to stopping Entware services, including TailScale.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Jul-23
+# Last Modified: 2024-Jul-24
 ###################################################################
 set -u
 
@@ -2080,7 +2080,7 @@ check_memory_and_prompt_reboot()
                 # Attempt to clear clears pagecache, dentries, and inodes after shutting down services
                 Say "Attempting to free up memory once more even more aggressively..."
 
-                # Stop Entware services before F/W flash #
+                # Stop Entware services to free some memory #
                 _EntwareServicesHandler_ stop
 
                 _ShutDownNonCriticalServices_
@@ -4198,7 +4198,7 @@ _Toggle_FW_UpdateCheckSetting_()
 }
 
 ##----------------------------------------##
-## Modified by Martinski W. [2024-Jun-28] ##
+## Modified by Martinski W. [2024-Jul-24] ##
 ##----------------------------------------##
 _EntwareServicesHandler_()
 {
@@ -4221,6 +4221,9 @@ _EntwareServicesHandler_()
           *) return 1 ;;
    esac
 
+   # Check if *NOT* skipping any services #
+   [ $# -gt 1 ] && [ "$2" = "-noskip" ] && skipServiceList=""
+
    _RenameSkippedService_()
    {
        [ -z "$skippedServiceList" ] && return 1
@@ -4239,8 +4242,8 @@ _EntwareServicesHandler_()
 
    [ -n "$serviceStr" ] && Say "Looking for Entware services..."
 
-   # Filter out services to skip and add a skip message #
-   if [ "$AllowVPN" = "ENABLED" ]
+   # Filter out services to skip and add a "skip message" #
+   if [ "$AllowVPN" = "ENABLED" ] && [ -n "$skipServiceList" ]
    then
       for skipService in $skipServiceList
       do
@@ -4801,9 +4804,6 @@ Please manually update to version $minimum_supported_version or higher to use th
     # Send last email notification before F/W flash #
     _SendEMailNotification_ START_FW_UPDATE_STATUS
 
-    # Stop Entware services before F/W flash #
-    _EntwareServicesHandler_ stop
-
     ##------------------------------------------##
     ## Modified by ExtremeFiretop [2024-Jun-30] ##
     ##------------------------------------------##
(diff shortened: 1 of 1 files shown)


ExtremeFiretop (approved):

ExtremeFiretop:
I'm wondering, with this solution, do we even have a need to keep tailscale alive anymore?

jksmurf:
I very much agree with Extremefiiretop here. Tailmon is set up to restart on reboot and force restart when it is not running, so as long as the MerlinAU process completes the FW update without Tailmon interfering, then having one less step where things can go wrong is a good idea..Simple is good.

Tailmon will (should) restart once the router reboots after the FW update, so the user will get access back (as long as the FW itself didn’t introduce other issues, but keeping a tailscale alive won’t fix those).k.

Martinski4GitHub:
> I'm wondering, with this solution, do we even have a need to keep tailscale alive anymore? Should we walk back/rollback that feature before we get too far along?
>
>
Currently, the call to stop the Entware services is made for 2 reasons:

**1)** To free some RAM in the case where the router is found in a "low memory" state when the script is checking for "Required vs Available" RAM to be able to continue with the "pre-flash" checks before doing the actual F/W flash.

**2)** Right before ejecting the USB-attached drive which is immediately before starting the actual F/W flash.

In case **1)** above, we do **not** want to stop TailScale because it's too early to terminate the connection so we do actually need to keep TailScale "alive" at this point during the "pre-flash" checks.

In case **2)** above, we're now stopping all services, no exceptions, so we can eject the USB-attached drive successfully.

ExtremeFiretop:
> > I'm wondering, with this solution, do we even have a need to keep tailscale alive anymore? Should we walk back/rollback that feature before we get too far along?
> 
> Currently, the call to stop the Entware services is made for 2 reasons:
> 
> **1)** To free some RAM in the case where the router is found in a "low memory" state when the script is checking for "Required vs Available" RAM to be able to continue with the "pre-flash" checks before doing the actual F/W flash.
> 
> **2)** Right before ejecting the USB-attached drive which is immediately before starting the actual F/W flash.
> 
> In case **1)** above, we do **not** want to stop TailScale because it's too early to terminate the connection so we do actually need to keep TailScale "alive" at this point during the "pre-flash" checks.
> 
> In case **2)** above, we're now stopping all services, no exceptions, so we can eject the USB-attached drive successfully.

@Martinski4GitHub .

I guess my thought is just remove the first Entware stop and stop everything at once. Only leave the second Entware stop. After the "trap" is called to ignore hangups.

That way, it's trapped, then we stop Entware services. Up until that point we can keep everything as it was, no?

Martinski4GitHub:
> > > I'm wondering, with this solution, do we even have a need to keep tailscale alive anymore? Should we walk back/rollback that feature before we get too far along?
> > 
> > 
> > Currently, the call to stop the Entware services is made for 2 reasons:
> > **1)** To free some RAM in the case where the router is found in a "low memory" state when the script is checking for "Required vs Available" RAM to be able to continue with the "pre-flash" checks before doing the actual F/W flash.
> > **2)** Right before ejecting the USB-attached drive which is immediately before starting the actual F/W flash.
> > In case **1)** above, we do **not** want to stop TailScale because it's too early to terminate the connection so we do actually need to keep TailScale "alive" at this point during the "pre-flash" checks.
> > In case **2)** above, we're now stopping all services, no exceptions, so we can eject the USB-attached drive successfully.
> 
> @Martinski4GitHub .
> 
> I guess my thought is just remove the first Entware stop and stop everything at once on the second step.. Only leave the second Entware stop after the "trap" is called to ignore hangups.
> 
> That way, it's trapped, then we stop Entware services. Up until that point we can keep everything as it was, no?
>
>
AFAIR, the main goal of the 1st call to stop Entware services was to try to free as much RAM as possible when the router is found to be below the "Required RAM" threshold. For routers with > 1GB RAM, or people with 1 or 2 services running and a couple of add-ons, having "Low RAM" may be an unlikely scenario, but some people have a lot of Entware services in addition to many add-ons running simultaneously. For such cases, I believe the 1st call is still valid and may still be useful.

P.S.
I have a work meeting in ~10 minutes. I'll take a look at the other messages and respond later in the evening. I expect a busy workday today.

ExtremeFiretop:
> > > > I'm wondering, with this solution, do we even have a need to keep tailscale alive anymore? Should we walk back/rollback that feature before we get too far along?
> > > 
> > > 
> > > Currently, the call to stop the Entware services is made for 2 reasons:
> > > **1)** To free some RAM in the case where the router is found in a "low memory" state when the script is checking for "Required vs Available" RAM to be able to continue with the "pre-flash" checks before doing the actual F/W flash.
> > > **2)** Right before ejecting the USB-attached drive which is immediately before starting the actual F/W flash.
> > > In case **1)** above, we do **not** want to stop TailScale because it's too early to terminate the connection so we do actually need to keep TailScale "alive" at this point during the "pre-flash" checks.
> > > In case **2)** above, we're now stopping all services, no exceptions, so we can eject the USB-attached drive successfully.
> > 
> > 
> > @Martinski4GitHub .
> > I guess my thought is just remove the first Entware stop and stop everything at once on the second step.. Only leave the second Entware stop after the "trap" is called to ignore hangups.
> > That way, it's trapped, then we stop Entware services. Up until that point we can keep everything as it was, no?
> 
> AFAIR, the main goal of the 1st call to stop Entware services was to try to free as much RAM as possible when the router is found to be below the "Required RAM" threshold. For routers with > 1GB RAM, or people with 1 or 2 services running and a couple of add-ons, having "Low RAM" may be an unlikely scenario, but some people have a lot of Entware services in addition to many add-ons running simultaneously. For such cases, I believe the 1st call is still valid and may still be useful.
> 
> P.S. I have a work meeting in ~10 minutes. I'll take a look at the other messages and respond later in the evening. I expect a busy workday today.

Ah yes, I see now, the second entware stop is at line **2143**, this is within the check_memory_and_prompt_reboot function.
NOT within the _RunFirmwareUpdateNow_ function.

Initially I thought you kept the entware stop at line **4873** and just added a "full stop" down below in the same function at line **4885**

ExtremeFiretop:
Now in hindsight, I see you actually removed the Entware stop call at  line 4873 and only moved it DOWN to line 4885, instead of duplicating it.

ExtremeFiretop:
We can ignore the remove the first entware stop line, I was confused when I wrote that.

>
>I guess my thought is just remove the first Entware stop and stop everything at once on the second step..

However, this still remains of my question:

>Only leave the second Entware stop after the "trap" is called to ignore hangups.
>
>That way, it's trapped, then we stop Entware services. Up until that point we can keep everything as it was, no?
>

Basically, in my head I thought read you had added a second within the same function, but we do not, we have 1 in each function as expected. But still my question remains, do we need to keep VPN active?

In my mind now, the answer is yes now. But in a case where the "check_memory_and_prompt_reboot" function runs, it will stop all Entware services EXCEPT tailscale, at line 2143:

```
                # Stop Entware services before F/W flash #
                _EntwareServicesHandler_ stop
```

If we did NOT include this feature, then theres potential for the memory function to randomly run, and boot you from your tailscale connection.

Martinski4GitHub:
> 
> In my mind now, the answer is yes now. But ONLY in a case where the "check_memory_and_prompt_reboot" function runs, since it will stop all Entware services EXCEPT tailscale, at line 2143:
> 
> ```
>                 # Stop Entware services before F/W flash #
>                 _EntwareServicesHandler_ stop
> ```
> 
> If we did NOT include this feature, then theres potential for the memory function to randomly run, and boot you from your tailscale connection, before reaching the "no-hang-up" step.
>
>
Exactly. When Entware services are stopped due to a "low RAM" check, Tailscale must remain "alive" still so the user can reach the final step just before the actual F/W flash starts.

ExtremeFiretop:
Synced with Gnuton in commit: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/186/commits/80e79aeeda1ea0e08bb12728b70001410d1555cf

PR #288, opened 2024-08-10
Remove Starting Set_FW_UpdateZIP_DirectoryPath

ExtremeFiretop:
Now that it's no longer a CLI switch/parameter and is instead a combination, we no longer need the Set_FW_UpdateZIP_DirectoryPath_ function at the start. 

They will already be in the advanced options menu, if they want to change the directory, they can use the function from the advanced options menu (option 1)

It was only originally included when the design was to be a CLI entry-point. 
(Feels a little weird now to have it as the starting function now)

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4833,44 +4833,38 @@ _RunOfflineUpdateNow_()
     offlineUpdateTrigger=true
     theMenuReturnPromptMsg="$advnMenuReturnPromptStr"
 
-    if _Set_FW_UpdateZIP_DirectoryPath_
+    clear
+    # Create directory for downloading & extracting firmware #
+    if ! _CreateDirectory_ "$FW_ZIP_DIR"
+    then
+        _ClearOfflineUpdateState_ 1 ; return 1
+    fi
+    printf "\n---------------------------------------------------\n"
+    printf "\nPlease copy your firmware ZIP file (using the *original* ZIP filename) to this directory:"
+    printf "\n[${GRNct}$FW_ZIP_DIR${NOct}]\n"
+    printf "\nPress '${GRNct}Y${NOct}' when completed, or '${REDct}N${NOct}' to cancel.\n"
+    printf "\n---------------------------------------------------\n"
+    if _WaitForYESorNO_
     then
         clear
-        # Create directory for downloading & extracting firmware #
-        if ! _CreateDirectory_ "$FW_ZIP_DIR"
-        then
-            _ClearOfflineUpdateState_ 1 ; return 1
-        fi
         printf "\n---------------------------------------------------\n"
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
@Martinski4GitHub 

This is just a suggestion for now, I don't really want to do another regular release just for this.
So if you approve this PR. Just approve it, don't merge it.

I'll do the change in the Gnuton branch instead.

Martinski4GitHub:
> Now that it's no longer a CLI switch/parameter and is instead a combination, we no longer need the Set_FW_UpdateZIP_DirectoryPath_ function at the start.
> 
> They will already be in the advanced options menu, if they want to change the directory, they can use the function from the advanced options menu (option 1)
> 
> It was only originally included when the design was to be a CLI entry-point. (Feels a little weird now to have it as the starting function now)
>
>
Yes, that makes sense and it's a very good point. Great catch!!

Martinski4GitHub:
> @Martinski4GitHub
> 
> This is just a suggestion for now, I don't really want to do another regular release just for this. So if you approve this PR. Just approve it, don't merge it.
> 
Yes, of course, there's no need for another release. This does not affect the regular update process anyway so it would affect only one user (that we know of ?? LOL!!). It's not an urgent fix either.

Martinski4GitHub (approved):
Approved & good to go.

ExtremeFiretop:
> > @Martinski4GitHub
> > This is just a suggestion for now, I don't really want to do another regular release just for this. So if you approve this PR. Just approve it, don't merge it.
> 
> Yes, of course, there's no need for another release. This does not affect the regular update process anyway so it would affect only one user (that we know of ?? LOL!!). It's not an urgent fix either.

Absolutely not, no regular user is impacted your correct on that front. It's not urgent or even required. The reason was that our user that helped with the feedback, tested it.

And was a bit confused by the starting prompt. He didn't understand why he needed to change the directory, (which he didn't) and then he tried to upload the firmware directly to the home/root folder instead of waiting until the next screen that said to upload the firmware now.

That's when I realized that entire screen isn't required anymore. If your in the advanced menu now, you do the combination, your expecting to see right away the directory to upload the zip, not the option 1 function to change the directory.

ExtremeFiretop:
I should mention that other than that, everything worked perfectly!

ExtremeFiretop:
> Approved & good to go.

Thank you for approving. I'll do the change in Gnuton now and close this PR. (As discussed)

ExtremeFiretop:
@Martinski4GitHub 

Updated in Gnuton commit: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/186/commits/91dbcb4cfacb6628596cb1d457c75ce17c69c6ac

(Will be included in the Gnuton merge)

PR #295, opened 2024-08-19
More improvements for detecting WebGUI access rules

Martinski4GitHub:
More code improvements to detect the following use cases of WebGUI access rules:

1) When the router's network IP address is not the commonly used base IP address (e.g. NOT "192.168.100.0").

2) When multiple CIDR IP address blocks are defined, we need to check if any include the router's LAN IP address.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Aug-16
+# Last Modified: 2024-Aug-18
 ###################################################################
 set -u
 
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
Approved and Merged! Great catch!

MerlinAU.sh
```diff
@@ -6477,11 +6547,6 @@ _SetSecondaryEMailAddress_()
    _WaitForEnterKey_ "$advnMenuReturnPromptStr"
 }
 
-# RegExp for IPv4 address #
-readonly IPv4octet_RegEx="([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])"
```
ExtremeFiretop, line 6481: I like the fact that you moved this up and made it a more universally accessible solution across the script. Nice job!

PR #305, opened 2024-10-02
Update Minimum Version

ExtremeFiretop:
Update Minimum Version, our minimum version is currently 6 versions behind production. Upping to 3 versions behind production. Last update in this regard was about 10 months ago.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -45,7 +45,7 @@ scriptUpdateNotify=0
 ##----------------------------------------##
 # For minimum supported firmware version check #
 MinFirmwareVerCheckFailed=false
-MinSupportedFirmwareVers="3004.386.12.0"
+MinSupportedFirmwareVers="3004.386.12.6"
 
 # For router model check #
 routerModelCheckFailed=false
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2024-Sep-19
+# Last Modified: 2024-Oct-02
 ###################################################################
 set -u
 
@@ -40,9 +40,9 @@ readonly high_risk_terms="factory default reset|features are disabled|break back
 DLRepoVersion=""
 scriptUpdateNotify=0
 
-##----------------------------------------##
-## Modified by Martinski W. [2024-Jul-29] ##
-##----------------------------------------##
+##------------------------------------------##
+## Modified by ExtremeFiretop [2024-Oct-02] ##
+##------------------------------------------##
 # For minimum supported firmware version check #
 MinFirmwareVerCheckFailed=false
 MinSupportedFirmwareVers="3004.386.12.6"


Martinski4GitHub:
> Update Minimum Version to 3004.386.12.6, our minimum version is currently 6 versions behind production or about 8 months behind production. Upping to 3 versions behind production. Last update in this regard was about 10 months ago.
>
>
With this new "Minimum Version" supported, users that happen to be running F/W versions below **3004.386.12.6** but above **3004.386.11.0** would not be able to update to the latest **3004.386.14.0** version.

What are the concerns if these users were still allowed to do the F/W update via MerlinAU?

Is the main concern due to the big jump/gap between the F/W versions?

ExtremeFiretop:
> 
> What are the concerns if these users were still allowed to do the F/W update via MerlinAU?
> 
> Is the main concern due to the big jump/gap between the F/W versions?

Exactly.

The intention of MerlinAU was always to keep a router up to date, not to be used to bring a router up to date over large version gaps. This was initially discussed in the forums when I was seeking feedback on the Addon. 

I've always wanted to keep fairly close minimum version due to this. 

The minimum version I see as dual purpose, it can be used when a specific firmware version is required for a successful update, but I don't believe that's happened recently, the second purpose is to be used to discourage users to install MerlinAU and use it to flash over many firmware versions.

Specific discussions comments were found here for reference:

Post in thread 'Seeking Feedback & Contributions: Merlin Auto Update Solutions' https://www.snbforums.com/threads/seeking-feedback-contributions-merlin-auto-update-solutions.87044/post-868605

Post in thread 'Seeking Feedback & Contributions: Merlin Auto Update Solutions' https://www.snbforums.com/threads/seeking-feedback-contributions-merlin-auto-update-solutions.87044/post-868608

Martinski4GitHub:
> > What are the concerns if these users were still allowed to do the F/W update via MerlinAU?
> > Is the main concern due to the big jump/gap between the F/W versions?
> 
> Exactly.
> 
> The intention of MerlinAU was always to keep a router up to date, not to be used to bring a router up to date over large version gaps. This was initially discussed in the forums when I was seeking feedback on the Addon.
> 
> I've always wanted to keep fairly close minimum version due to this.
> 
> The minimum version I see as dual purpose, it can be used when a specific firmware version is required for a successful update, but I don't believe that's happened recently, the second purpose is to be used to discourage users to install MerlinAU and use it to do one time flashes over many firmware versions.
> 
>
Yeah, I see. If that's the case, then what about the **3004.388.xx** version series?
There's no "minimum version" check for it even though the first version came out on Dec-2022. 
IOW, a user running F/W version **3004.388.1.0** can still update to the latest **3004.388.8_2** (9 versions later, ~21 months apart). Should there be a minimum check for **3004.388.xx** as well?

ExtremeFiretop:
> > > What are the concerns if these users were still allowed to do the F/W update via MerlinAU?
> > > Is the main concern due to the big jump/gap between the F/W versions?
> > 
> > 
> > Exactly.
> > The intention of MerlinAU was always to keep a router up to date, not to be used to bring a router up to date over large version gaps. This was initially discussed in the forums when I was seeking feedback on the Addon.
> > I've always wanted to keep fairly close minimum version due to this.
> > The minimum version I see as dual purpose, it can be used when a specific firmware version is required for a successful update, but I don't believe that's happened recently, the second purpose is to be used to discourage users to install MerlinAU and use it to do one time flashes over many firmware versions.
> 
> Yeah, I see. If that's the case, then what about the **3004.388.xx** version series? There's no "minimum version" check for it even though the first version came out on Dec-2022. IOW, a user running F/W version **3004.388.1.0** can still update to the latest **3004.388.8_2** (9 versions later, ~21 months apart). Should there be a minimum check for **3004.388.xx** as well?

I never considered separating the branch versions with their own checks, but it's not a bad idea at all.

I just didn't want to bump to 388 when we still support two AC routers on 386 technically... But if we split it up then we could have a minimum for each... I like the idea the more I think about it.

Martinski4GitHub:
> > > > What are the concerns if these users were still allowed to do the F/W update via MerlinAU?
> > > > Is the main concern due to the big jump/gap between the F/W versions?
> > > 
> > > 
> > > Exactly.
> > > The intention of MerlinAU was always to keep a router up to date, not to be used to bring a router up to date over large version gaps. This was initially discussed in the forums when I was seeking feedback on the Addon.
> > > I've always wanted to keep fairly close minimum version due to this.
> > > The minimum version I see as dual purpose, it can be used when a specific firmware version is required for a successful update, but I don't believe that's happened recently, the second purpose is to be used to discourage users to install MerlinAU and use it to do one time flashes over many firmware versions.
> > 
> > 
> > Yeah, I see. If that's the case, then what about the **3004.388.xx** version series? There's no "minimum version" check for it even though the first version came out on Dec-2022. IOW, a user running F/W version **3004.388.1.0** can still update to the latest **3004.388.8_2** (9 versions later, ~21 months apart). Should there be a minimum check for **3004.388.xx** as well?
> 
> I never considered separating the branch versions with their own checks, but it's not a bad idea at all.
> 
> I just didn't want to bump to 388 when we still support two AC routers on 386 technically... But if we split it up then we could have a minimum for each... I like the idea the more I think about it.
>
>
I think at some point, we'll have to "split" the current 3 branches (**3004.386**.x, **3004.388**.x & **3006.102**.x) in order to run a "minimum version" check for each, as appropriate for the router, but taking into account that for some routers it might be possible to go from one branch to the next (e.g. from **3004.388**.x to **3006.102**.x). Something to consider... :>)

PR #367, opened 2024-11-26
Fixes: Cron Schedule Calculations, Changelog Verification, Post-Update Version Comparison

ExtremeFiretop:
Fixes: 
1. Cron Schedule Calculations as identified by YOU! HEHE!
2. Changelog Verification as identified by ME!
3. Post-Update Version Comparison with Gnuton failing, as identified by ME!

Also includes 2 additional features:

1. Includes the changelog in the update notification email as requested by @nlurker here: https://www.snbforums.com/threads/merlinau-v1-3-6-the-ultimate-firmware-auto-updater-gnuton-support.91326/post-935218
2. All caps risk phrases when using Plain Text email as requested by @TheS1R here: https://www.snbforums.com/threads/merlinau-v1-3-6-the-ultimate-firmware-auto-updater-gnuton-support.91326/post-934807

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -5658,25 +5658,6 @@ _ChangelogVerificationCheck_()
             _DoCleanUp_
             return 1
         else
-            formatted_release_version=$(echo "$release_version" | awk -F. '{
-                if ($1 ~ /^[0-9]{4}$/) {  # Check for a four-digit prefix
-                    if (NF == 4 && $4 == "0") {
-                        printf "%s.%s", $2, $3  # For version like 3004.388.5.0, remove the last .0
-                    } else if (NF == 4) {
-                        printf "%s.%s.%s", $2, $3, $4  # For version like 3004.388.5.2, keep the last digit
-                    }
-                } else if (NF == 3) {  # For version without a four-digit prefix
-                    if ($3 == "0") {
-                        printf "%s.%s", $1, $2  # For version like 388.5.0, remove the last .0
-                    } else {
-                        printf "%s.%s.%s", $1, $2, $3  # For version like 388.5.2, keep the last digit
-                    }
-                }
-            }')
-
-            # Define regex patterns for both versions
-            release_version_regex="[0-9]{4}\.${formatted_release_version//./[._]}\s*\([0-9]{1,2}-[A-Za-z]+-[0-9]{4}\)"
-
             # Use awk to format the version based on the number of initial digits
             formatted_current_version=$(echo "$current_version" | awk -F. '{
                 if ($1 ~ /^[0-9]{4}$/) {  # Check for a four-digit prefix
@@ -5698,6 +5679,24 @@ _ChangelogVerificationCheck_()
                 }
             }')
 
+            formatted_release_version=$(echo "$release_version" | awk -F. '{
+                if ($1 ~ /^[0-9]{4}$/) {  # Check for a four-digit prefix
+                    if (NF == 4 && $4 == "0") {
+                        printf "%s.%s", $2, $3  # For version like 3004.388.5.0, remove the last .0
+                    } else if (NF == 4) {
+                        printf "%s.%s.%s", $2, $3, $4  # For version like 3004.388.5.2, keep the last digit
+                    }
+                } else if (NF == 3) {  # For version without a four-digit prefix
+                    if ($3 == "0") {
+                        printf "%s.%s", $1, $2  # For version like 388.5.0, remove the last .0
+                    } else {
+                        printf "%s.%s.%s", $1, $2, $3  # For version like 388.5.2, keep the last digit
+                    }
+                }
+            }')
+
+            # Define regex patterns for both versions
+            release_version_regex="[0-9]{4}\.${formatted_release_version//./[._]}\s*\([0-9]{1,2}-[A-Za-z]+-[0-9]{4}\)"
             current_version_regex="[0-9]{4}\.${formatted_current_version//./[._]}\s*\([0-9]{1,2}-[A-Za-z]+-[0-9]{4}\)"
 
             if "$isGNUtonFW"
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -3060,6 +3060,7 @@ _GetNodeInfo_()
     node_build_name="Unreachable"
     node_lan_hostname="Unreachable"
     node_label_mac="Unreachable"
+    NodeGNUtonFW=false
 
     ## Check for Login Credentials ##
     credsBase64="$(Get_Custom_Setting credentials_base64)"
@@ -3122,6 +3123,9 @@ _GetNodeInfo_()
     node_lan_hostname="$(echo "$htmlContent" | grep -o '"lan_hostname":"[^"]*' | sed 's/"lan_hostname":"//')"
     node_label_mac="$(echo "$htmlContent" | grep -o '"label_mac":"[^"]*' | sed 's/"label_mac":"//')"
 
+    # Check if installed F/W NVRAM vars contain "gnuton" #
+    if echo "$node_extendno" | grep -iq "gnuton"
+    then NodeGNUtonFW=true ; fi
     # Combine extracted information into one string #
     Node_combinedVer="${node_firmver}.${node_buildno}.$node_extendno"
 
@@ -5992,6 +5996,23 @@ _CheckNodeFWUpdateNotification_()
              echo ""
              echo "AiMesh Node <b>${nodefriendlyname}</b> with MAC address <b>${node_label_mac}</b> requires update from <b>${1}</b> to <b>${2}</b> version."
              echo "(<b>${1}</b> --> <b>${2}</b>)"
+             echo "Please click here to review the latest changelog:"
+             if "$NodeGNUtonFW"
+             then
+                 Gnuton_changelogurl="$(GetLatestChangelogUrl "$FW_GITURL_RELEASE")"
+                 echo "$Gnuton_changelogurl"
+             else
+                 if [ "$node_firmver" -eq 3006 ]
+                 then
+                     changeLogURL="${CL_URL_3006}"
+                 elif echo "$node_firmver" | grep -qE "^386[.]"
+                 then
+                     changeLogURL="${CL_URL_386}"
+                 else
+                     changeLogURL="${CL_URL_NG}"
+                 fi
+                 echo "$changeLogURL"
+             fi
              echo "Automated update will be scheduled <b>only if</b> MerlinAU is installed on the node."
            } > "$tempNodeEMailList"
        fi
@@ -6007,6 +6028,23 @@ _CheckNodeFWUpdateNotification_()
          echo ""
          echo "AiMesh Node <b>${nodefriendlyname}</b> with MAC address <b>${node_label_mac}</b> requires update from <b>${1}</b> to <b>${2}</b> version."
          echo "(<b>${1}</b> --> <b>${2}</b>)"
+             echo "Please click here to review the latest changelog:"
+             if "$NodeGNUtonFW"
+             then
+                 Gnuton_changelogurl="$(GetLatestChangelogUrl "$FW_GITURL_RELEASE")"
+                 echo "$Gnuton_changelogurl"
+             else
+                 if [ "$node_firmver" -eq 3006 ]
+                 then
+                     changeLogURL="${CL_URL_3006}"
+                 elif echo "$node_firmver" | grep -qE "^386[.]"
+                 then
+                     changeLogURL="${CL_URL_386}"
+                 else
+                     changeLogURL="${CL_URL_NG}"
+                 fi
+                 echo "$changeLogURL"
+             fi
          echo "Automated update will be scheduled <b>only if</b> MerlinAU is installed on the node."
        } > "$tempNodeEMailList"
    fi
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -5700,8 +5700,8 @@ _ChangelogVerificationCheck_()
             }')
 
             # Define regex patterns for both versions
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
@Martinski4GitHub 

I am really happy with my work here, feel free to review, test and merge whenever your available :)

Martinski4GitHub:
> @Martinski4GitHub
> 
> I am really happy with my work here, feel free to review, test and merge whenever your available :)
>
Reviewing the PR now. Looking good so far...

MerlinAU.sh
```diff
@@ -3985,18 +4018,29 @@ _IncrementDay_()
...
     local curr_dom="$1"
     local dom_expr="$2"
-    local domStart  domEnd
+    local domStart domEnd expanded_days
 
     if [ "$dom_expr" = "*" ]
     then  # Matches any day of the month #
         return 0
+    elif echo "$dom_expr" | grep -q '/'
+    then
```
Martinski4GitHub, line 4034: Ah, so the specific cases for "day intervals" were missing altogether. LOL!!!
Like pretty much all users, we don't set up the cron schedule with day intervals within the week or the month so never noticed at all that the cases were missing, but somebody out there will try someday so now they will work fine!!
Good job, bud!!

ExtremeFiretop: @Martinski4GitHub 

Exactly! A very clear oversight on my part, but a very great catch on your part!

MerlinAU.sh
```diff
@@ -5640,8 +5700,8 @@ _ChangelogVerificationCheck_()
             }')
 
             # Define regex patterns for both versions
-            release_version_regex="$formatted_release_version \([0-9]{1,2}-[A-Za-z]{3}-[0-9]{4}\)"
-            current_version_regex="$formatted_current_version \([0-9]{1,2}-[A-Za-z]{3}-[0-9]{4}\)"
+            release_version_regex="[0-9]{4}\.${formatted_release_version//./[._]}\s*\([0-9]{1,2}-[A-Za-z]+-[0-9]{4}\)"
+            current_version_regex="[0-9]{4}\.${formatted_current_version//./[._]}\s*\([0-9]{1,2}-[A-Za-z]+-[0-9]{4}\)"
```
Martinski4GitHub, line 5704: These regular expressions will not work for the "**Changelog-386.txt**" file because they're looking for 4 digits **before** the version string but they don't exist. You need to account for both cases: **with** or **without** the 4-digit prefix.

ASUSWRT-Merlin_ChangeLog-386

ExtremeFiretop: Ahhhh, good call. I didn't check that changelog, I only checked NG and 3006.

I'm assuming if I simply remove the requirement for the 3004 prefix. That it will work correctly? Could you test that?

ExtremeFiretop: Just sitting back down from after dinner now. Give me a few moments to test the change or NG and 3006 as well

Martinski4GitHub: > Ahhhh, good call. I didn't check that changelog, I only checked NG and 3006.
> 
> I'm assuming if I simply remove the requirement for the 3004 prefix. That it will work correctly? Could you test that?
>
>
Yeah, that's one way to do it. However, IMO, the better solution would be to fix the code above where all the 4-digit prefixes (i.e. "**3004**" & "**3006**") seem to get removed from the "formatted" version strings.

I think the key is to remove it **only** for the "**386**" builds but do not remove for the "**388**" & "**3006.102**" builds. Once you do that, the regular expression become much simpler.

My 2 cents.

ExtremeFiretop: I think that was initially done exactly because of the 386 changelog. We only really needed to match the specific minor and patch version without caring about the prefix since the prefix is assumed within that specific changelog (3006, NG, etc)

Martinski4GitHub: > Just sitting back down from after dinner now. Give me a few moments to test the change or NG and 3006 as well
>
I'll be "summoned for dinner" in about 10 minutes here... Just FYI. :>)

ExtremeFiretop: The date regex matching is to make sure we don't match specific versions within the notes if I remember correctly.

Martinski4GitHub: > I think that was initially done exactly because of the 386 changelog. We only really needed to match the specific minor and patch version without caring about the prefix since the prefix is assumed within that specific changelog (3006, NG, etc)
>
>
So previously the 4-digit prefix was kept **except** for the "386" builds?
If so, why was it changed? I don't see a good reason to have changed it...

Martinski4GitHub: > The date regex matching is to make sure we don't match specific versions within the notes if I remember correctly.
>
Oh, yeah, there's nothing wrong with that part of the regex. We want to match **ONLY** the strings that have the versions PLUS the dates in one line.

ExtremeFiretop: > > I think that was initially done exactly because of the 386 changelog. We only really needed to match the specific minor and patch version without caring about the prefix since the prefix is assumed within that specific changelog (3006, NG, etc)
> 
> So previously the 4-digit prefix was kept **except** for the "386" builds? If so, why was it changed? I don't see a good reason to have changed it...

No I'm saying the code above removes the prefix since the beginning. It was "initially" done as in the original design removed the prefix for that 386 changelog. It's not included in the changelog. So we removed it from the current and upcoming version strings for comparison against the changelogs.

The prefix is not required for the 3004 or 3006 changelogs since the prefix is assumed already within those changelogs.

ExtremeFiretop: > The "**3004**" & "**3006**" prefixes seem to get removed by the `awk` code above even though the specific prefix **is** **required** for the "**388**" & "**102**" builds to search for the "version + date" in the changelog.
> 

Is it required though? My understanding of the fix is me removing the suffix from the code above, the regex was just being adjusted to be more forgiving with the _ or . And the spacing around the date. I added the prefix because I only checked NG and 3006 but I don't think it's required if wasn't in the code previously.

ExtremeFiretop: In my test here: it doesn't appear to be required:

image

ExtremeFiretop: Checkout commit: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/367/commits/2c70b4c40bd2223bdd85d804f235ce46ae59dd62

That was my suggestion, I don't think the prefix is required in those changelogs for results to be matched.

ExtremeFiretop: Btw I'm not trying to be dismissive of your idea to modify the top code above the regex to only drop the prefix for 386. 

I'm just trying to understand why keep the prefix at all if it's not required for the matching to work?

I only initially added the prefix match to the regex because I made the regex more forgiving and loser matching with the rear of the version (swapping dots or underscores) so I decided to tighten up the front end of the version number to balance out the change and not make the regex too lose. 

But really the date matching is good enough and the prefix doesn't appear to be required so it can just be dropped from the regex and now it works with all changelogs again.

The two real fixes here are the dropping of the suffix in the code above for the current version, and making the regex loser with the dots and underscore matching since some changelogs use _2 and others might use .2

Martinski4GitHub: > Btw I'm not trying to be dismissive of your idea to modify the top code above the regex to only drop the prefix for 386.
> 
> I'm just trying to understand why keep the prefix at all if it's not required for the matching to work?
> 
> I only initially added the prefix match to the regex because I made the regex more forgiving and loser matching with the rear of the version (swapping dots or underscores) so I decided to tighten up the front end of the version number to balance out the change and not make the regex too lose.
> 
> But really the date matching is good enough and the prefix doesn't appear to be required so it can just be dropped from the regex and now it works with all changelogs again.
> 
> The two real fixes here are the dropping of the suffix in the code above for the current version, and making the regex loser with the dots and underscore matching since some changelogs use _2 and others might use .2 and being more forgiving on the amount of spacing between the date and the version.
>
>

I believe this is a case where we have differences of opinion WRT "proper implementation" because we're operating under a different set of coding rules & standards. Let me explain a little. Pretty much all rules & standards I've worked with dictate that whenever we're trying to find a matching line, we always try to search for as much of the expected string as possible, from the beginning to the end, and make exceptions to the rule only for very special circumstances.

For the particular case of the changelogs, my default operating mode is to match as much of the specific version string as possible followed by the date string. This means that I would try to match the full version, including the 4-digit prefix (i.e. "3004" or "3006") unless there's a special case (i.e. "386" builds). Therefore, in my view, the 4-digit prefixes are required for proper line matching, taking into account exceptions for special case(s), because we're looking for precise matches.

Again, this is how I always operate in my professional work, and using these coding rules & standards for my implementations spills over to all extracurricular projects & hobbies I collaborate in.

Now, I know there are "many ways to skin a cat." And we obviously have different coding backgrounds & standards, so I understand your point of view even though I would have done it differently to match my own working operating mode.

I hope the above clarifies where I'm coming from WRT my coding recommendations for solutions.

Ultimately, if your fixes are working well for all test cases, then we're good to go.

Martinski4GitHub (approved):
Good to go!!

Martinski4GitHub:
> Good to go!!
>
FYI, I'd like to run the latest script version through the Linter tool. Lately, a lot of code was added and/or modified so I want to check again to make sure we're still "clean." ;>)

MerlinAU.sh
ExtremeFiretop: Yes that clarifies what you mean by it being required. Because in your standards it is and I'm not familiar with that.

But at a technical level when I tested the fix it didn't seem to physically be required which is why the original implementation from the start dropped the prefixes.

Now I think I'm understanding you want to change it so everything uses the prefixes by default? And only for builds detected as 386 do we do the matching by dropping the prefix?

I think I understand the viewpoint, I just was tagging a fix with minimal changes required, and already I overstepped by adding the prefix match and not thinking / double checking the 386 changelogs.

Martinski4GitHub: > Yes that clarifies what you mean by it being required. Because in your standards it is and I'm not familiar with that.
> 
> But at a technical level when I tested the fix it didn't seem to physically be required which is why the original implementation from the start dropped the prefixes.
> 
> Now I think I'm understanding you want to change it so everything uses the prefixes by default? And only for builds detected as 386 do we do the matching by dropping the prefix?
> 
> I think I understand the viewpoint, I just was tagging/thinking of a fix with minimal changes required, and already I overstepped by adding the prefix match and not thinking / double checking the 386 changelogs.]
>
No worries, we're good to go for this round of fixes. If in the future any bugs are found in this area again, we can revisit the implementation.

Let me connect to my company's server and then my workstation to run the Linter tool. I'll let you know of the results.

ExtremeFiretop:
> > Good to go!!
> 
> FYI, I'd like to run the latest script version through the Linter tool. Lately, a lot of code was added and/or modified so I want to check again to make sure we're still "clean." ;>)

Fair point! Do let me know the results before we send off the next release!

Btw as you mentioned in the review comment. If something comes up again we can change the implementation to something else. I'll let you lead the implementation next time. I've seen you do the same thing for the changelogs selection, I even used that code for the emails to decide which changelog to send, so I guess we would use similar logic for the changelog matching to decide when to strip the prefix?

It wasn't something I considered at the time, and it's why I was so confused when you said you wanted to add them. Because in my test "removing" them was the universal fix.

But actually, even if something doesn't come up, If your confident that implementation is better I'm willing to just hand it over to you to fix up. I'm good at quick fixes, but remember I'm not anywhere near your level, I'm more somewhere between you and a newbie, on the dirty and quick fix level hahahaha 🤣 

I know Viktor joked about it before, but I'm pretty serious when I say it 😆

Martinski4GitHub:
> > Good to go!!
> 
> FYI, I'd like to run the latest script version through the Linter tool. Lately, a lot of code was added and/or modified so I want to check again to make sure we're still "clean." ;>)
>
>
I forgot they told us this morning. Today at 12:05 AM, the company's servers are being taken offline, one by one, for maintenance purposes (i.e. kernel updates/patches, security updates, etc.) so I got the email just at midnight reminding us.

Anyway, I'll do it tomorrow evening.

Have a good night, bud.

ExtremeFiretop:
> > > Good to go!!
> > 
> > 
> > FYI, I'd like to run the latest script version through the Linter tool. Lately, a lot of code was added and/or modified so I want to check again to make sure we're still "clean." ;>)
> 
> I forgot they told us this morning. Today at 12:05 AM, the company's servers are being taken offline, one by one, for maintenance purposes (i.e. kernel updates/patches, security updates, etc.) so I got the email just at midnight reminding us.
> 
> Anyway, I'll do it tomorrow evening.
> 
> Have a good night, bud.

Have a goodnight Bud!

Martinski4GitHub:
> > > Good to go!!
> > 
> > 
> > FYI, I'd like to run the latest script version through the Linter tool. Lately, a lot of code was added and/or modified so I want to check again to make sure we're still "clean." ;>)
> 
> Fair point! Do let me know the results before we send off the next release!
> 
> Btw as you mentioned in the review comment. If something comes up again we can change the implementation to something else. I'll let you lead the implementation next time. I've seen you do the same thing for the changelogs selection, I even used that code for the emails to decide which changelog to send, so I guess we would use similar logic for the changelog matching to decide when to strip the prefix?
> 
> It wasn't something I considered at the time, and it's why I was so confused when you said you wanted to add them. Because in my test "removing" them was the universal fix.
> 
> But actually, even if something doesn't come up, If your confident that implementation is better I'm willing to just hand it over to you to fix up. I'm good at quick fixes, but remember I'm not anywhere near your level, I'm more somewhere between you and a newbie, on the dirty and quick fix level hahahaha 🤣
> 
>
I won't have a lot of time in the next 10 working days or so, in addition to family gatherings/celebrations (i.e. Thanksgiving, two birthdays, etc.), given my work deadlines with the final release due on Dec-18th, I'll be pretty busy. Perhaps after that, we can revisit this if nothing else more critical comes up.

PR #377, opened 2024-12-30
Minor UI Improvement 

ExtremeFiretop:
Small Improvements

In the same spirit as your last PR: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/376
I have made 2 small adjustments for dev.

Visually the Format type for the core shell script is redundant with the option of "Email Format Type" displaying the format. It simply looks better I think:

image

Also walked back a change from my previous PR here: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/367 , and instead decided to specifically select to remove the Gnuton branch, else we get this with offline flashing of Alphas:

image

So there is a reason we are being picky after all and not removing all suffixes, and it's related to the alphas and betas which I didn't consider the first time.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -1686,8 +1686,8 @@ _CreateEMailContent_()
         fwNewUpdateVersion="$(Get_Custom_Setting "FW_New_Update_Notification_Vers")"
    fi
 
-   #Remove any suffix starting with "-" or "_" to avoid version comparison failures#
-   fwInstalledVersion="$(echo "$fwInstalledVersion" | sed -E 's/[-_].*$//')"
+   # Remove "_rog" or "_tuf" or -gHASHVALUES or -Gnuton* suffix to avoid version comparison failure, can't remove all for proper beta and alpha comparison #
+   fwInstalledVersion="$(echo "$fwInstalledVersion" | sed -E 's/(_(rog|tuf)|-g[0-9a-f]{10}|-gnuton[0-9]+)$//')"
 
    case "$1" in
        FW_UPDATE_TEST_EMAIL)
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
> Small Improvements
> 
> In the same spirit as your last PR: #376 I have made 2 small adjustments for dev.
> 
> Visually the Format type for the core shell script is redundant with the option of "Email Format Type" displaying the format. It simply looks better I think:
> 
>
Good call!!

> 
> Also walked back a change from my previous PR here: #367 , and instead decided to specifically select to remove the Gnuton branch, else we get this with offline flashing of Alphas:
>
> ...
> 
> So there is a reason we are being picky after all and not removing all suffixes, and it's related to the alphas and betas which I didn't consider the first time.
>
Ah yes, good catch!!

MerlinAU.sh
```diff
@@ -8722,7 +8722,7 @@ _ShowAdvancedOptionsMenu_()
        fi
        if "$sendEMailNotificationsFlag"
        then
-           printf "\n${padStr}[Currently ${GRNct}ENABLED${NOct}, Format: ${GRNct}${sendEMailFormaType}${NOct}]\n"
```
Martinski4GitHub, line 8725: The closing right square bracket char is missing!
```
printf "\n${padStr}[Currently ${GRNct}ENABLED${NOct}]\n"
                                                   ^^^
```

ExtremeFiretop: This is why you should never do an in-place edit on your phone! :P The other PR has it properly

ExtremeFiretop: @Martinski4GitHub 

Fixed :) (using inplace edit on my phone)
Do what I say, not what I do? Is that how it goes? LOL

Martinski4GitHub (approved):
Good to go!!

PR #393, opened 2025-01-20
Modified Uninstallation Behavior

Martinski4GitHub:
Modified Uninstallation to NOT keep/save the configuration file by default when called from the WebGUI.


ExtremeFiretop (approved):

ExtremeFiretop:
Tested and confirmed working exactly as promised.
This is more or less the expected behavior based on the messaging anyways, but we can of course, modify this in the future (maybe give the user the selection) as you mentioned yesterday

PR #397, opened 2025-01-28
Fixes to handle AMTM Email Config File NOT found

Martinski4GitHub:
Made fixes & improvements to handle the scenario where the AMTM email config file is NOT found so that all email notification parameters must be shown as disabled on the WebGUI page.

### DIFF
--- a/MerlinAU.asp
+++ b/MerlinAU.asp
@@ -29,7 +29,7 @@
 <script language="JavaScript" type="text/javascript">
 
 /**----------------------------**/
-/** Last Modified: 2025-Jan-26 **/
+/** Last Modified: 2025-Jan-27 **/
 /** Intended for 1.4.0 Release **/
 /**----------------------------**/
 
@@ -55,6 +55,16 @@ const InvREDct = '<span style="margin-left:4px; background-color:#C81927; color:
 const InvYLWct = '<span style="margin-left:4px; background-color:yellow; color:black;">&nbsp;'
 const InvCYNct = '<span style="margin-left:4px; background-color:cyan; color:black;">&nbsp;'
 const InvCLEAR = '&nbsp;</span>'
+
+/**----------------------------------------**/
+/** Modified by Martinski W. [2025-Jan-27] **/
+/**----------------------------------------**/
+var externalCheckID = 0x00;
+var externalCheckOK = true;
+var externalCheckMsg = '';
+var isEMailConfigEnabledInAMTM = false;
+var scriptAutoUpdateCronSchedHR = 'TBD';
+var fwAutoUpdateCheckCronSchedHR = 'TBD';
 
 /**-------------------------------------**/
 /** Added by Martinski W. [2025-Jan-13] **/
@@ -737,21 +747,8 @@ function ValidateDirectoryPath (formField)
    }
 }
 
-/**-------------------------------------**/
-/** Added by Martinski W. [2025-Jan-16] **/
-/**-------------------------------------**/
-var externalCheckID = 0x00;
-var externalCheckOK = true;
-var externalCheckMsg = '';
-
-/**-------------------------------------**/
-/** Added by Martinski W. [2025-Jan-20] **/
-/**-------------------------------------**/
-var scriptAutoUpdateCronSchedHR = 'TBD';
-var fwAutoUpdateCheckCronSchedHR = 'TBD';
-
 /**----------------------------------------**/
-/** Modified by Martinski W. [2025-Jan-20] **/
+/** Modified by Martinski W. [2025-Jan-27] **/
 /**----------------------------------------**/
 function GetExternalCheckResults()
 {
@@ -766,6 +763,7 @@ function GetExternalCheckResults()
         {
             document.getElementById('Script_AutoUpdate_SchedText').textContent = 'Schedule: '+scriptAutoUpdateCronSchedHR;
             document.getElementById('FW_AutoUpdate_CheckSchedText').textContent = 'Schedule: '+fwAutoUpdateCheckCronSchedHR;
+            SetUpEmailNotificationFields();
 
             // Skip during form submission //
             if (isFormSubmitting) { return true ; }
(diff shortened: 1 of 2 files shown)


Martinski4GitHub:
@ExtremeFiretop,

This PR now handles the issue you found WRT the AMTM config file **not** found & disabling the associated email notification parameters on the WebGUI page. Here are some screenshots showing 3 different scenarios:

1) AMTM config file **not** set up or **not** found:

MerlinAU_WebGUI_EmailNotify#1

2) AMTM config file OK and "Email Notifications" option is **disabled**:

MerlinAU_WebGUI_EmailNotify#2

3) AMTM config file OK and "Email Notifications" option is **enabled**:

MerlinAU_WebGUI_EmailNotify#3

Please take test & validate when you have some time.

ExtremeFiretop:
> @ExtremeFiretop,
> 
> This PR now handles the issue you found WRT the AMTM config file **not** found & disabling the associated email notification parameters on the WebGUI page. Here are some screenshots showing 3 different scenarios:
> 
> 1. AMTM config file **not** set up or **not** found:
> 
> MerlinAU_WebGUI_EmailNotify#1
> 
> 2. AMTM config file OK and "Email Notifications" option is **disabled**:
> 
> MerlinAU_WebGUI_EmailNotify#2
> 
> 3. AMTM config file OK and "Email Notifications" option is **enabled**:
> 
> MerlinAU_WebGUI_EmailNotify#3
> 
> Please take test & validate when you have some time.

I like the distinction between the AMTM config file not found, and found but option is disabled.
My original logic didn't account for this I don't think... Will test and advise back shortly!

ExtremeFiretop (approved):

ExtremeFiretop:
@Martinski4GitHub 

Oh yeah, you got the magic my friend, works exactly as expected now:

Email function missing:

image

And email function configured but disabled:

image

MerlinAU.asp
```diff
@@ -56,6 +56,16 @@ const InvYLWct = '<span style="margin-left:4px; background-color:yellow; color:b
 const InvCYNct = '<span style="margin-left:4px; background-color:cyan; color:black;">&nbsp;'
 const InvCLEAR = '&nbsp;</span>'
 
+/**----------------------------------------**/
+/** Modified by Martinski W. [2025-Jan-27] **/
+/**----------------------------------------**/
+var externalCheckID = 0x00;
```
ExtremeFiretop, line 62: For educational purposes, can you explain the flow for the CheckHelper.js?

My understanding is that this is a bridge/hook for the WebUI and script

So basically we have 3 methods to send data?

1. Whatever is in www/user is /available to the WebUI to read (like our configuration file)
2. The shared settings file in jffs/addons which allows us to export data from the WebUI
3. This helper.js which allows us to do what that the other two don't exactly? It seems to be setting up key variables and default states that some other code will rely on? (Outside of the configuration file?)

ExtremeFiretop:
@Martinski4GitHub 

Merged! Works as advertised! And resolves the reported issues/bug!

MerlinAU.asp
Martinski4GitHub: > For educational purposes, can you explain the logic flow for the CheckHelper.js?
> 
> My understanding is that this is a bridge/hook for the WebUI and script
> 
> So basically we have 3 methods to send data?
> 
>     1. Whatever is in www/user is /available to the WebUI to read (like our configuration file)
> 
>     2. The shared settings file in jffs/addons which allows us to export data from the WebUI
> 
>     3. This helper.js in our addons directory, which allows us to do what that the other two don't exactly? It seems to be setting up key variables and default states that some other code will rely on? (Outside of the configuration file?)
>
>
Yes, the helper JS file is a one-way "hook" method that allows us to dynamically define JS variables and re-assign them any data or states that we want, based on some trigger event or state change that the shell script wants to communicate to the WebGUI. So the JS variables are first initialized in the regular JS source file, and later on their modified values are communicated via the helper JS file which gets loaded into the regular JS code via the **GetExternalCheckResults()** function. At that point, the redefined & modified variables can be used just like any other regular JS vars.

Technically, one could achieve the same results via the configuration file, but often times we don't want to use the config file to hold such data or states which can be much more easily communicated via the helper JS file (e.g. the validation results of the "F/W Update Directory" or the human-readable strings of the current cron schedules).

HTH.

ExtremeFiretop: > > For educational purposes, can you explain the logic flow for the CheckHelper.js?
> > My understanding is that this is a bridge/hook for the WebUI and script
> > So basically we have 3 methods to send data?
> > ```
> > 1. Whatever is in www/user is /available to the WebUI to read (like our configuration file)
> > 
> > 2. The shared settings file in jffs/addons which allows us to export data from the WebUI
> > 
> > 3. This helper.js in our addons directory, which allows us to do what that the other two don't exactly? It seems to be setting up key variables and default states that some other code will rely on? (Outside of the configuration file?)
> > ```
> 
> Yes, the helper JS file is a one-way "hook" method that allows us to dynamically define JS variables and re-assign them any data or states that we want, based on some trigger event or state change that the shell script wants to communicate to the WebGUI. So the JS variables are first initialized in the regular JS source file, and later on their modified values are communicated via the helper JS file which gets loaded into the regular JS code via the **GetExternalCheckResults()** function. At that point, the redefined & modified variables can be used just like any other regular JS vars.
> 
> Technically, one could achieve the same results via the configuration file, but often times we don't want to use the config file to hold such data or states which can be much more easily communicated via the helper JS file (e.g. the validation results of the "F/W Update Directory" or the human-readable strings of the current cron schedules).
> 
> HTH.

This is absolutely perfect. Thank you for taking the time it helps shape a fuller picture of how you intend to use this so I don't mess it up.

I had a gut feeling that was more or less it's purpose but the clarif6goes a long way, and I agree that not all things we want to hold in the configuration file, it was one of my thoughts when I started this project honestly is I didn't want to dirty our configuration file with webUI specific configurations

PR #401, opened 2025-02-12
More Build Type Refinements

ExtremeFiretop:
@Martinski4GitHub 

Dropping more of the firmware specific logic for the TUF and ROG builds; I think in theory this is the last of the build type changes. If you notice or think or anything else; feel free :)


Martinski4GitHub (approved):
Looks good!

PR #409, opened 2025-02-18
Improved SHA256 Signature Extraction

Martinski4GitHub:
Code improvements for extracting SHA256 signatures.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2025-Feb-16
+# Last Modified: 2025-Feb-17
 ###################################################################
 set -u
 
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
@ExtremeFiretop,

I took a closer look at the fix because I was wondering why the following line was now needed in the modified `curl` cmd call:
```
sed -e 's/SHA256 signatures:Latest release://'
```
It turned out it was because you changed the syntax (the dots were removed from the previous call) here:
```
sed -n '/<pre[^>]*>/,/<\/pre>/p'  |
sed -e 's/<[^>]*>//g'
```
After doing testing with the original code, I found that the root cause of the problem was not really with the `curl` cmd; the issue was due to the `cut` cmd used later to extract the signature. The fix would be to use `awk` instead of the `cut` cmd here:
```
dl_sig="$(echo "$checksums" | grep "$(basename "$firmware_file")" | awk -F ' ' '{print $1}')"
```
In the end, the better fix would be to adjust the `curl` and use `awk` cmd for extraction.

I have submitted a PR to each branch with the resulting fixes.
Please review when you get some time.

ExtremeFiretop:
> @ExtremeFiretop,
> 
> I took a closer look at the fix because I was wondering why the following line was now needed in the modified `curl` cmd call:
> 
> ```
> sed -e 's/SHA256 signatures:Latest release://'
> ```
> 
> It turned out it was because you changed the syntax (the dots were removed from the previous call) here:
> 
> ```
> sed -n '/<pre[^>]*>/,/<\/pre>/p'  |
> sed -e 's/<[^>]*>//g'
> ```
> 
> After doing testing with the original code, I found that the root cause of the problem was not really with the `curl` cmd; the issue was due to the `cut` cmd used later to extract the signature. The fix would be to use `awk` instead of the `cut` cmd here:
> 
> ```
> dl_sig="$(echo "$checksums" | grep "$(basename "$firmware_file")" | awk -F ' ' '{print $1}')"
> ```
> 
> In the end, the better fix would be to adjust the `curl` and use `awk` cmd for extraction.
> 
> I have submitted a PR to each branch with the resulting fixes. Please review when you get some time.

Great! Thanks for double-checking and finding that the cut command was the culprit. 
It's always helpful to have another pair of eyes on these kinds of issues. 

I really appreciate the time you took to review and verify the fix!

Martinski4GitHub:
> > @ExtremeFiretop,
> > I took a closer look at the fix because I was wondering why the following line was now needed in the modified `curl` cmd call:
> > ```
> > sed -e 's/SHA256 signatures:Latest release://'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It turned out it was because you changed the syntax (the dots were removed from the previous call) here:
> > ```
> > sed -n '/<pre[^>]*>/,/<\/pre>/p'  |
> > sed -e 's/<[^>]*>//g'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After doing testing with the original code, I found that the root cause of the problem was not really with the `curl` cmd; the issue was due to the `cut` cmd used later to extract the signature. The fix would be to use `awk` instead of the `cut` cmd here:
> > ```
> > dl_sig="$(echo "$checksums" | grep "$(basename "$firmware_file")" | awk -F ' ' '{print $1}')"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In the end, the better fix would be to adjust the `curl` and use `awk` cmd for extraction.
> > I have submitted a PR to each branch with the resulting fixes. Please review when you get some time.
> 
> Great! Thanks for double-checking and finding that the cut command was the culprit. It's always helpful to have another pair of eyes on these kinds of issues.
> 
> I really appreciate the time you took to review and verify the fix!
>
>
Of course, bud. We got each other's back when reviewing our PRs. It's a team effort always!! :>).

Have a good night, bud. Sleep tight!!

PR #418, opened 2025-03-01
Fix The Update Check Status Post Upgrade

ExtremeFiretop:
Fix The Update Check Status Post Upgrade in the WebUI

Issue seems to be it does not check or set the **FW_UPDATE_CHECK** value in the settings file post upgrade as seen below:

image

We should be seeing this run:

image

Which sets the value.

Solution seems to be to simply add the missing: _ConfirmCronJobForFWAutoUpdates_ within the _DoInstallation_ routine.
Without this, on first upgrade after running: `sh /jffs/scripts/MerlinAU.sh develop` 

We get this result in the configuration file:

image

Simply running the script a second time also fixes the "bug" but we can add this to the Install routine I think without any impact.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -14,6 +14,10 @@ readonly SCRIPT_NAME="MerlinAU"
 ## Set to "master" for Production Releases ##
 SCRIPT_BRANCH="dev"
 
+##FOR TESTING/DEBUG ONLY##
+if true ; then SCRIPT_BRANCH="WebFun" ; fi
+##FOR TESTING/DEBUG ONLY##
+
 ##----------------------------------------##
 ## Modified by Martinski W. [2024-Jul-03] ##
 ##----------------------------------------##
@@ -8901,6 +8905,7 @@ _DoStartupInit_()
        _AutoStartupHook_ create 2>/dev/null
        _AutoServiceEvent_ create 2>/dev/null
    fi
+
 }
 
 ##----------------------------------------##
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
> Fix The Update Check Status Post Upgrade in the WebUI
> 
> Issue seems to be it does not check or set the **FW_UPDATE_CHECK** value in the settings file post upgrade as seen below:
> ...
> 
> We should be seeing this run:
> ....
> 
> Which sets the value.
> 
> Solution seems to be to simply add the missing: _ConfirmCronJobForFWAutoUpdates_ within the _DoInstallation_ routine. 
>
>
Yeah, that should work just fine.

>
> Without this, on first upgrade after running: `sh /jffs/scripts/MerlinAU.sh develop`
> 
> We get this result in the configuration file:
> 
> 
> Simply running the script a second time also fixes the "bug" but we can add this to the Install routine I think without any impact.
>
Sounds good.

Martinski4GitHub (approved):
Good to go!!

ExtremeFiretop:
> > Fix The Update Check Status Post Upgrade in the WebUI
> > Issue seems to be it does not check or set the **FW_UPDATE_CHECK** value in the settings file post upgrade as seen below:
> > ...
> > We should be seeing this run:
> > ....
> > Which sets the value.
> > Solution seems to be to simply add the missing: _ConfirmCronJobForFWAutoUpdates_ within the _DoInstallation_ routine.
> 
> Yeah, that should work just fine.
> 
> > Without this, on first upgrade after running: `sh /jffs/scripts/MerlinAU.sh develop`
> > We get this result in the configuration file:
> > Simply running the script a second time also fixes the "bug" but we can add this to the Install routine I think without any impact.
> 
> Sounds good.

Now that's in dev; I'll just test it again! :)

ExtremeFiretop:
@Martinski4GitHub 

Bingo! Was his name-o! I now see the missing step when doing the update process:

image

And loading the WebUI right away shows "Enabled"

image

Thanks buddy for the review!

PR #419, opened 2025-03-08
WebGUI Functionality Improvement

Martinski4GitHub:
Added functionality to the WebGUI so that users can optionally select to run a "Router Login Test" when saving the password to verify that the login credentials are indeed valid. This is similar to the option available in the CLI menu where users can select to run the same router login test.

### DIFF
--- a/MerlinAU.asp
+++ b/MerlinAU.asp
@@ -29,8 +29,8 @@
 <script language="JavaScript" type="text/javascript">
 
 /**----------------------------**/
-/** Last Modified: 2025-Mar-01 **/
-/** Intended for 1.4.0 Release **/
+/** Last Modified: 2025-Mar-07 **/
+/** Intended for 1.4.x Release **/
 /**----------------------------**/
 
 // Separate variables for shared and AJAX settings //
@@ -825,7 +825,7 @@ function ValidateDirectoryPath (formField, dirType)
 function GetExternalCheckResults()
 {
     $.ajax({
-        url: '/ext/MerlinAU/CheckHelper.js',
+        url: '/ext/MerlinAU/checkHelper.js',
         dataType: 'script',
         timeout: 5000,
         error: function(xhr){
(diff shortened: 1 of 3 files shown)


Martinski4GitHub:
@ExtremeFiretop,

Whenever you have time, Please review and validate the new functionality added with this PR.

Here is the general workflow:

1) When users first start the WebGUI, they'll see an "**UNKNOWN**" status for the password field because the code doesn't know whether the password is valid (even if the user knows it is in fact valid).

MerlinAU_v1 4 0 _WebGUI_Unknown

2) The user can select the option to "**Run login test when saving password**" by clicking on the checkbox below the password field (default is always **unchecked**). Upon selection, the user gets an alert pop-up window explaining that the login test will log them out of the WebGUI, so the user must confirm the selection.

MerlinAU_v1 4 0 _WebGUI_Notice

3) If the "Router Login Test" checkbox is selected when the user clicks on the "**Save**" button for that section, the normal process starts but after ~10 secs, the user is automatically logged out of the WebGUI.

4) When the user logs back into the WebGUI, they are presented with 2 possible status results:
    
     a) Password in **INVALID** because the login test failed:

MerlinAU_v1 4 0 _WebGUI_Error

MerlinAU_v1 4 0 _WebGUI_Invalid

     b) Password in VERIFIED because the login test was successful.

MerlinAU_v1 4 0 _WebGUI_Verified

5) OTOH, if the user does **NOT** select to run the "Router Login Test" **before** saving the password, the normal "Save" process occurs and when completed, the user will see the "**NOT verified**" status for the password field. This status will remain until the user selects to run the Router Login Test to verify the credentials.

MerlinAU_v1 4 0 _WebGUI_NotVerified

6) Here's the tooltip showing the username that is used with the password to run the login test. This is useful in cases where sometimes the username value gets corrupted in NVRAM.

MerlinAU_v1 4 0 _WebGUI_Tooltip

Let me know if this new functionality works well for you.

ExtremeFiretop (approved):

ExtremeFiretop:
Good to go!

Absolutely beautiful @Martinski4GitHub !!

Love the additon!

ExtremeFiretop:
image

PR #395, opened 2025-01-23
New Checkmarks on WebGUI

Martinski4GitHub:
Modified WebGUI page with checkmarks to give users the option to "Keep configuration file" when uninstalling, and to "Bypass postponed days" when doing a F/W update check.

MerlinAU_WebGUI_Checkmarks

The default value for each checkmark is always **false**/**disabled** so users must choose to enable them every time they want to perform the desired action.

PR #423, opened 2025-03-11
Minor Fixes

ExtremeFiretop:
Minor Fixes

1. Added a check to validate the downloaded firmware version by MerlinAU is infact the version found by the router's webs_update.sh script.

Seems to me what happened at some point for my RT-AX92U router is @GNUton built the firmwares; (lets say at midnight) and his Github action automatically published it. Shortly after it was automatically published, the Webs-Update.sh script ran and found the update.

MerlinAU then ran a few hours later by cron job, by that point the builds were hidden. (Gnuton had unpublished or hidden the builds by 3AM) But because they were published and the webs_update.sh script detected it, and started to flash.

So instead MerlinAU downloaded the latest 388.8_2 and downgraded me from the latest alpha lol. 

image
image

This additional check will cancel the script if this happens again and the versions don't match. Along with additional Gnuton specific changes to make sure we grab the right version by parsing for matching version and matching Product ID.

2. Fixed the post email upgrade weirdness we found in January here: 
https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/395#issuecomment-2614229361

image

In short; we shouldn't be using: `_GetLatestFWUpdateVersionFromRouter_ 1` because that does not split on the dash. That was a me mistake when I implemented this I think?

While `_GetLatestFWUpdateVersionFromRouter_`  does split the version information at the dash for things like rog, or -gnuton.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4545,29 +4545,52 @@ _GetLatestFWUpdateVersionFromGithub_()
         search_type="pure\|squashfs\|ubi"
     fi
 
+    if ! "$offlineUpdateTrigger"
+    then
+        # Get the router version from the router itself
+        local router_version="$(_GetLatestFWUpdateVersionFromRouter_ 1)"
+    else
+        # Get the router version from the router itself
+        local router_version="$(Get_Custom_Setting "FW_New_Update_Notification_Vers")"
+    fi
+
+    if [ -z "$router_version" ]; then
+        echo "**ERROR** **NO_ROUTER_VERSION**"
+        return 1
+    fi
+
     # Fetch the latest release data from GitHub #
     local release_data="$(curl -s "$url")"
 
     # Construct the grep pattern based on search_type #
     local grep_pattern="\"browser_download_url\": \".*${PRODUCT_ID}.*\(${search_type}\).*\.\(w\|pkgtb\)\""
 
-    # Filter the JSON for the desired firmware using grep and head to fetch the URL
-    local download_url="$(echo "$release_data" | 
-        grep -o "$grep_pattern" | 
-        grep -o "https://[^ ]*\.\(w\|pkgtb\)" | 
-        head -1)"
+    # Extract all matched download URLs
+    local download_urls="$(echo "$release_data" | \
+        grep -o "$grep_pattern" | \
+        grep -o "https://[^ ]*\.\(w\|pkgtb\)")"
 
     # Check if a URL was found
-    if [ -z "$download_url" ]
+    if [ -z "$download_urls" ]
     then
         echo "**ERROR** **NO_GITHUB_URL**"
         return 1
     else
-        # Extract the version from the download URL or release data
-        local version="$(echo "$download_url" | grep -oE "${PRODUCT_ID}[_-][0-9.]+[^/]*" | sed "s/${PRODUCT_ID}[_-]//;s/.w$//;s/_/./g")"
-        echo "$version"
-        echo "$download_url"
-        return 0
+        # Loop through each matching URL and compare version to router_version
+        local url_item version
+        for url_item in $download_urls; do
+            # Extract the version portion from the URL
+            local version="$(echo "$url_item" \
+                | grep -oE "${PRODUCT_ID}_[^ ]*\.(w|pkgtb)" \
+                | sed "s/${PRODUCT_ID}_//;s/.w$//;s/.pkgtb$//;s/.ubi$//;s/_/./g" | head -n1)"
+
+            # If this URL’s version matches the router version, we're done
+            if [ "$version" = "$router_version" ]; then
+                echo "$version"
+                echo "$url_item"
+                return 0
+            fi
+        done
     fi
 }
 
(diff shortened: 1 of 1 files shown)


MerlinAU.sh
```diff
@@ -2846,7 +2846,7 @@ _CreateEMailContent_()
    fwInstalledVersion="$(_GetCurrentFWInstalledLongVersion_)"
    if ! "$offlineUpdateTrigger"
    then
-        fwNewUpdateVersion="$(_GetLatestFWUpdateVersionFromRouter_ 1)"
+        fwNewUpdateVersion="$(_GetLatestFWUpdateVersionFromRouter_)"
```
ExtremeFiretop, line 2849: This is the post upgrade email fix. Simple and easy

MerlinAU.sh
```diff
@@ -4545,29 +4545,52 @@ _GetLatestFWUpdateVersionFromGithub_()
         search_type="pure\|squashfs\|ubi"
     fi
 
+    if ! "$offlineUpdateTrigger"
```
ExtremeFiretop, line 4548: The rest below is to fix the issue of not validating the downloaded version is the correct version we got notified about from Webs_update.sh

MerlinAU.sh
```diff
@@ -8384,6 +8433,16 @@ Please manually update to version ${GRNct}${MinSupportedFirmwareVers}${NOct} or
         fi
     fi
 
+    if ! "$offlineUpdateTrigger"
```
ExtremeFiretop, line 8436: Finally; add a final check before flashing to validate it all and cancel if they still don't match somehow.
The above code is Gnuton specific, this catch is for all firmwares, Gnuton or Merlin

MerlinAU.sh
```diff
@@ -4545,29 +4545,52 @@ _GetLatestFWUpdateVersionFromGithub_()
...
-        echo "$version"
-        echo "$download_url"
-        return 0
+        # Loop through each matching URL and compare version to router_version
+        local url_item version
+        for url_item in $download_urls; do
+            # Extract the version portion from the URL
+            local version="$(echo "$url_item" \
+                | grep -oE "${PRODUCT_ID}_[^ ]*\.(w|pkgtb)" \
+                | sed "s/${PRODUCT_ID}_//;s/.w$//;s/.pkgtb$//;s/.ubi$//;s/_/./g" | head -n1)"
```
ExtremeFiretop, line 4585: With the head -n1 here, I was getting multiple lines with the existing code:

```
Admin@GT-BE98_Pro:/tmp/home/root# echo "$version"
4-gnuton1
3004.388.8.4-gnuton1.puresqubi
```

TheS1R: > Without the head -n1 here, I was getting multiple lines with the existing code:
> 
> ```
> Admin@GT-BE98_Pro:/tmp/home/root# echo "$version"
> 4-gnuton1
> 3004.388.8.4-gnuton1.puresqubi
> ```

That darn gnuton is continuing to give you grief!

ExtremeFiretop: > > Without the head -n1 here, I was getting multiple lines with the existing code:
> > ```
> > Admin@GT-BE98_Pro:/tmp/home/root# echo "$version"
> > 4-gnuton1
> > 3004.388.8.4-gnuton1.puresqubi
> > ```
> 
> That darn gnuton is continuing to give you grief!

We'll get there. 

It's a little bit different since it's dealing with GitHub instead of Sourceforge and they don't package them the same, and they have their own UIs (TUF instead of ROG) and so on,  it's always fun to identify spots to improve. 

Both these things were on my to do list for a while but I never got around to it, and no one reported anything hahaha

ExtremeFiretop: This is probably one of the only add-ons you'll find where the logic flow takes a completely different route between Merlin and Gnuton, usually scripts just run on Gnuton with little to no changes.

But because we are dealing with flashing the firmware automatically that does have a different process between them haha

Martinski4GitHub:
> Minor Fixes
> 
>     1. Added a check to validate the downloaded firmware version by MerlinAU is infact the version found by the router's webs_update.sh script.
> 
> 
> Seems to me what happened at some point for my RT-AX92U router is @gnuton built the firmwares; (lets say at midnight) and his Github action automatically published it. Shortly after it was automatically published, the Webs-Update.sh script ran and found the update.
> 
> MerlinAU then ran a few hours later by cron job, by that point the builds were hidden. (Gnuton had unpublished or hidden the builds by 3AM) But because they were published and the webs_update.sh script detected it, and started to flash.
> 
> So instead MerlinAU downloaded the latest 388.8_2 and downgraded me from the latest alpha lol.
> 
> ...
> 
> This additional check will cancel the script if this happens again and the versions don't match. Along with additional Gnuton specific changes to make sure we grab the right version by parsing for matching version and matching Product ID.
> 
>     2. Fixed the post email upgrade weirdness we found in January here:
>        [New Checkmarks on WebGUI #395 (comment)](https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/395#issuecomment-2614229361)
> 
> ....
> 
> In short; we shouldn't be using: `_GetLatestFWUpdateVersionFromRouter_ 1` because that does not split on the dash. That was a me mistake when I implemented this I think?
> 
> While `_GetLatestFWUpdateVersionFromRouter_` does split the version information at the dash for things like rog, or -gnuton.
>
>
Very good finds, bud!! I can only assume that this scenario with a Gnuton F/W build being published and then, hours later, being "recalled" (i.e. unpublished or hidden somehow) is an extremely rare event, and that's why we've never seen a user reporting that MerlinAU "downgraded" the installed F/W version. I don't recall reading about this same scenario happening with RMerlin's F/W releases at all.

In any case, it's always good to add code that double-checks & verifies that the correct F/W release version is in fact being installed. Excellent job, bud!

Martinski4GitHub (approved):
Good to go!!

ExtremeFiretop:
> > Minor Fixes
> > ```
> > 1. Added a check to validate the downloaded firmware version by MerlinAU is infact the version found by the router's webs_update.sh script.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Seems to me what happened at some point for my RT-AX92U router is @gnuton built the firmwares; (lets say at midnight) and his Github action automatically published it. Shortly after it was automatically published, the Webs-Update.sh script ran and found the update.
> > MerlinAU then ran a few hours later by cron job, by that point the builds were hidden. (Gnuton had unpublished or hidden the builds by 3AM) But because they were published and the webs_update.sh script detected it, and started to flash.
> > So instead MerlinAU downloaded the latest 388.8_2 and downgraded me from the latest alpha lol.
> > ...
> > This additional check will cancel the script if this happens again and the versions don't match. Along with additional Gnuton specific changes to make sure we grab the right version by parsing for matching version and matching Product ID.
> > ```
> > 2. Fixed the post email upgrade weirdness we found in January here:
> >    [New Checkmarks on WebGUI #395 (comment)](https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/395#issuecomment-2614229361)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ....
> > In short; we shouldn't be using: `_GetLatestFWUpdateVersionFromRouter_ 1` because that does not split on the dash. That was a me mistake when I implemented this I think?
> > While `_GetLatestFWUpdateVersionFromRouter_` does split the version information at the dash for things like rog, or -gnuton.
> 
> Very good finds, bud!! I can only assume that this scenario with a Gnuton F/W build being published and then, hours later, being "recalled" (i.e. unpublished or hidden somehow) is an extremely rare event, and that's why we've never seen a user reporting that MerlinAU "downgraded" the installed F/W version. I don't recall reading about this same scenario happening with RMerlin's F/W releases at all.
> 

Thank you! Yes i think it's fairly rare, but we should probably build in a "safety" against it. I won't lie, it really confused me to wake up to my router DOWNGRADED by MerlinAU hahaha

> In any case, it's always good to add code that double-checks & verifies that the correct F/W release version is in fact being installed. 

Agreed!

ExtremeFiretop:
Btw check SNB, spdMerlin reports. I'm troubleshooting as we speak

PR #442, opened 2025-04-07
Update version.txt

Martinski4GitHub:
Removed build "timestamp" to make sure those users currently still on older versions (e.g. **1.3.10** or **1.3.11**) don't run into issues when trying to upgrade to the latest **1.4.0** version.


ExtremeFiretop (approved):

ExtremeFiretop:
> Removed build "timestamp" to make sure those users currently still on older versions (e.g. **1.3.10** or **1.3.11**) don't run into issues when trying to upgrade to the latest **1.4.0** version.

So removing it ended up being the better choice than updating it I guess.
So be it, as live we learn. im surprised people are that far behind but it's the way of things.

ExtremeFiretop:
The good news is removing it should only matter for this release I guess

Martinski4GitHub:
> > Removed build "timestamp" to make sure those users currently still on older versions (e.g. **1.3.10** or **1.3.11**) don't run into issues when trying to upgrade to the latest **1.4.0** version.
> 
> So removing it ended up being the better choice than updating it I guess. So be it, we live we learn. 
>
>
Yeah, the error message may not show up because the "old" code checking for the version might be failing silently when finding the "timestamp" and not recognizing it for what it is.

>
>
> im surprised people are that far behind but it's the way of things.
>
>
Some users fall behind because they prefer (for whatever reason 🤷🏻‍♂) not to enable the "Automatic Script Update" option, and there are those who don't frequent the SNB Forums regularly or often enough to realize that there are more recent updated versions. Also, in my experience, most users don't read carefully the Release Notes.

ExtremeFiretop:
> > The good news is removing it should only matter for this release I guess
> 
> Yeah, once all users are on the **1.4.0** production release, it will not matter whether the timestamp is present.

We'll give it more time before we "forget" about it while pushing to production.
I was thinking of updating the workflow to remove it, if it exists, when pushing to production, but I'm not sure there'd much value there, as we discussed this should be really only until everyone is on 1.4.0

Martinski4GitHub:
> > > The good news is removing it should only matter for this release I guess
> > 
> > 
> > Yeah, once all users are on the **1.4.0** production release, it will not matter whether the timestamp is present.
> 
> We'll give it more time before we "forget" about it while pushing to production. I was thinking of updating the workflow to remove it, if it exists, when pushing to production, but I'm not sure there'd much value there, as we discussed this should only really matter until everyone is on 1.4.0
>
>
Exactly. This is a one-off scenario when migrating to the 1.4.0 release. From then on, it will be "business as usual."

PR #445, opened 2025-04-08
Fixes & Improvements

Martinski4GitHub:
1) Modified code to check and make sure the WebGUI ASP file is downloaded and installed.

2) Modified code to check and make sure all configuration default settings are initialized.


Martinski4GitHub:
@ExtremeFiretop,

This PR takes care of the 2 root causes that were generating the issues reported on the SNB Forums.
Please review and test whenever you get a chance. I have tested this PR on 3 separate routers by performing a fresh install on each one.

MerlinAU.sh
```diff
@@ -176,14 +176,6 @@ keepConfigFile=false
 bypassPostponedDays=false
 runLoginCredentialsTest=false
 
-##---------------------------------------##
```
ExtremeFiretop, line 179: Are we removing these because we will lean on the PR to thelonelycoder instead?

MerlinAU.sh
```diff
@@ -5086,23 +5160,23 @@ _Toggle_VPN_Access_()
         printf "Disabling this feature will shut down Tailscale/ZeroTier VPN access during updates.\n"
         printf "Proceed if you do not need remote VPN access during firmware updates.\n"
 
-        if _WaitForYESorNO_ "\nProceed to ${GRNct}DISABLE${NOct}?"
+        if _WaitForYESorNO_ "\nProceed to ${REDct}DISABLE${NOct}?"
```
ExtremeFiretop, line 5163: I actually liked the way it was here. I think is this exactly how I meant it originally as I explained to the user.
I would probably just add a quick note to the description about how the default selection is Disabled as the user requested.

MerlinAU.sh
```diff
@@ -9237,21 +9312,24 @@ _DoStartupInit_()
...
    _CreateDirPaths_
-   _InitCustomSettingsConfig_
+   _InitCustomDefaultsConfig_
+   _InitCustomUserSettings_
    _CreateSymLinks_
    _InitHelperJSFile_
    _SetVersionSharedSettings_ local "$SCRIPT_VERSION"
    _SetVersionSharedSettings_ server "$SCRIPT_VERSION"
    _DownloadScriptFiles_ install
+   _CheckAndSetBackupOption_
```
ExtremeFiretop, line 9331: I see you added them here, that makes sense to me.

MerlinAU.sh
```diff
@@ -10702,15 +10780,13 @@ _DoInitializationStartup_()
        _DoExit_ 1
    fi
 
-   _CheckAndSetBackupOption_
```
ExtremeFiretop, line 10705: What I'm not following is why it appeared to work for me if I moved out from behind the "install" parameter check.

MerlinAU.sh
Martinski4GitHub: > Are we removing this because we will lean on the PR to thelonelycoder instead?
>
>
No, the PR submitted to AMTM may take a while before getting merged into the master branch. So I modified the code so that the call with the "**install**" parameter is explicitly made when checking for the ASP file, and it's **not** found mounted or installed. This handles the problem at its root while we wait for the AMTM pull request to percolate and before it reaches the end users.

MerlinAU.sh
Martinski4GitHub: > I actually liked the way it was here. I think is this exactly how I meant it originally as I explained to the user. I would probably just add a quick note to the description about how the default selection is Disabled as the user requested.
>
>
When I went to check the WebGUI page, the "**VPN Access**" setting follows the same color pattern used for all others: "Disabled" in **Red** and "Enabled" is **Green**. So I think we really need to make this option consistent with all the rest. We either change the WebGUI or the CLI menu, and it was simpler and easier to make the change in the CLI menu. But I have no real preference; either way works for me.

MerlinAU.sh
Martinski4GitHub: > What I'm not following is why it appeared to work for me if I moved out from behind the "install" parameter check.
>
Do you mean moving the function calls *before* checking for the "install" parameter?
If so, those calls worked when moved because the `_DoInitializationStartup_()` function is called early during the initialization stage, and the calls were **not** done by the `_DoInstallation_()` function, where they should have been called - but now they are.

MerlinAU.sh
ExtremeFiretop: Okay this may just be one of those agree to disagree there.

I would argue the CLI and WebUI elements have no connection and are two distinct elements. Similar to the argument you made about my symmetry on the WebUI in the past when I tried to make all the boxes the same size. The argument was that they were 2 different sections and their element didn't need to match.

I would say the CLI and the WebUI also doesn't need to match and functionally it made sense. But I wont make a deal about it I don't care enough about colors.

Martinski4GitHub: > Okay this may just be one of those agree to disagree there.
> 
> I would argue the CLI and WebUI elements have no connection and are two distinct elements. Similar to the argument you made about my symmetry on the WebUI in the past when I tried to make all the boxes the same size. The argument was that they were 2 different sections and their element didn't need to match.
> 
> I would say the CLI and the WebUI also doesn't need to match and functionally it made sense. But I wont make a deal about it I don't care enough about colors.
>
>
Understood. We'll agree to disagree.

MerlinAU.sh
ExtremeFiretop: > > What I'm not following is why it appeared to work for me if I moved out from behind the "install" parameter check.
> 
> Do you mean moving the function calls _before_ checking for the "install" parameter? If so, those calls worked when moved because the `_DoInitializationStartup_()` function is called early during the initialization stage, and the calls were **not** done by the `_DoInstallation_()` function, where they should have been called - but now they are.

Yes exactly what I mean. And Okay, reviewing the code I think I follow the logic. So instead of moving it up in the initialization, we added it to the do install steps which j did notice above in another review note.

PR #447, opened 2025-04-08
Check for Version File

Martinski4GitHub:
Added a check for the version.txt file.


ExtremeFiretop (approved):

ExtremeFiretop:
> @ExtremeFiretop,
> 
> Can you re-test on the AiMesh node using this latest PR?

Completed find below results:

image

And:

image

It did add the version.txt. but only after the update check was already completed, it makes the flow feel a little weird 
(notified that theres an update available when there isn't, and mentioned the cron jobs twice)

I'm wondering if we moved it up further in the process so it completed before the update check if that would be better.
To refernece; my original set install method worked like this from my PR:

image

ExtremeFiretop:
@Martinski4GitHub 

Any value in just reverting back to that method? but keeping your changes for the configuration file?

ExtremeFiretop:
Tell the snow to go away. View from our my window right now. The spring needs to stop teasing us

IMG20250408022423

Martinski4GitHub:
> > @ExtremeFiretop,
> > Can you re-test on the AiMesh node using this latest PR?
> 
> Completed find below results:
> ...
> And:
> ...
> 
> It did add the version.txt. but only after the update check was already completed, it makes the flow feel a little weird (notified that theres an update available when there isn't, and mentioned the cron jobs twice)
> 
> I'm wondering if we moved it up further in the process so it completed before the update check if that would be better. To refernece; my original set install method worked like this from my PR:
> 
Yes, good point. I've moved up the function call. Can you try again?

Martinski4GitHub:
> @Martinski4GitHub
> 
> Any value in just reverting back to that method? but keeping your changes for the configuration file?
>
>
The issue with that method, as written, is that it will keep making the "install" parameter call every single time on AiMesh nodes. Why? Because it checks for an entry in the hook file that will never be present on AiMesh nodes.

ExtremeFiretop:
> > @Martinski4GitHub
> > Any value in just reverting back to that method? but keeping your changes for the configuration file?
> 
> The issue with that method, as written, is that it will keep making the "install" parameter call every single time on AiMesh nodes. Why? Because it checks for an entry in the hook file that will never be present on AiMesh nodes.

On the AiMesh nodes you would have this "#MerlinAU" tag:

image

ExtremeFiretop:
I guess thats dependent on the update checks being enabled though?
Idk it seemed like a working solution for me

ExtremeFiretop:
Oh I see; the tag name would need to be changed for that to match as well.

Martinski4GitHub:
> > > @Martinski4GitHub
> > > Any value in just reverting back to that method? but keeping your changes for the configuration file?
> > 
> > 
> > The issue with that method, as written, is that it will keep making the "install" parameter call every single time on AiMesh nodes. Why? Because it checks for an entry in the hook file that will never be present on AiMesh nodes.
> 
> On the AiMesh nodes you would have this "#MerlinAU" tag:
> 
Yeah, but it may not be always there, and the code was looking for the "**#MerlinAU#**" tag

Martinski4GitHub:
> I guess thats dependent on the update checks being enabled though? 
>
Correct, which would make the method unreliable. I'd rather make the code check for exactly the file(s) that we are looking for and expecting to be installed.

PR #449, opened 2025-04-08
Patch Update Process

ExtremeFiretop:
Patch Update Process


ExtremeFiretop:
To resolve the discussed issues from tonight:

This screen is wrong for a timestamp update:

image

It should be this one:

image

And this is wrong; it should restart the script when completed to reflect the new version.

image

PR #452, opened 2025-04-10
Code Improvements

Martinski4GitHub:
Some additional code improvements.

- Make sure the "**FW_New_Update_Changelog_Approval**" option is initialized properly under all circumstances.
- Show the Password "**Verified**" status in GREEN = GOOD.
- Minor cleanup.


Martinski4GitHub:
@ExtremeFiretop.

Linter tool came back with a very good report: Found no errors and no new warnings.

Sample screenshot showing the Password "verified" status in green:

MerlinAU_v1 4 0 _CLI_PswdOK

MerlinAU.sh
```diff
@@ -4408,15 +4435,12 @@ _GetLoginCredentials_()
         printf "Encoded Credentials:\n"
         printf "${GRNct}$loginCredsENC${NOct}\n"
 
-        # Prompt to test the credentials #
         if _WaitForYESorNO_ "\nWould you like to test the current login credentials?"
         then
             _TestLoginCredentials_ "$loginCredsENC" || continue
-        else
-            _UpdateLoginPswdCheckHelper_ UNKNOWN
```
Martinski4GitHub, line 4416: This call to set the password status to UNKNOWN is unnecessary. 
The goal is to leave the previously check status alone. If it was already verified and is still the same, we don't want to set it to UNKNOWN just because the user chooses not to run the password check again.
Also, the function checks for that scenario, so no need to make the call to UNKNOWN.

ExtremeFiretop:
All good to go!

I like the idea to turn the text green if it's verified. It looks good and less flashy to the eyes 😂

Martinski4GitHub:
> All good to go!
> 
> I like the idea to turn the text green if it's verified. It looks good and less flashy to the eyes 😂
>
>
Yeah, I think so, and it fits our expectations of green color indicating a "**good**" status. Other colors (e.g. red, purple, yellow) instinctively trigger our notion of "Is something missing?" or "Did something go wrong?" despite what the words say.

PR #453, opened 2025-04-11
Patch Changelog and Email Notifications and Fix Uninstall

ExtremeFiretop:
Patch Changelog and Email Notifications

PR #454, opened 2025-04-11
Fix and Improvements

Martinski4GitHub:
- Fixed a regression bug introduced in PR #453, which was allowing installation in all unsupported models.
- Miscellaneous code improvements.


ExtremeFiretop:
Actually definitely not a bug.

By design. Has it get installed to be uninstalled.
Thanks.

MerlinAU.sh
```diff
@@ -10832,14 +10837,22 @@ _MainMenu_()
...
+## Modified by Martinski W. [2025-Apr-11] ##
+##----------------------------------------##
 _DoInitializationStartup_()
 {
+   local theParam=""
+   if [ $# -gt 0 ] && [ -n "$1" ] ; then theParam="$1" ; fi
 
-   if [ $# -gt 0 ] && [ -n "$1" ] && \
-      echo "$1" | grep -qE "^(install|startup)$"
+   if ! _CheckForMinimumRequirements_ && [ "$theParam" != "uninstall" ]
```
ExtremeFiretop, line 10848: It works exactly as originally designed in my originally PR without a need for extra variables.
I'll give you an opportunity to explain your thinking why any of this is required when the original design works.

Otherwise feel free to close the PR.

ExtremeFiretop:
@Martinski4GitHub 

I believe you think the unsupported models should stop the install?
Heck no, sign day one it allowed uninstall but blocked updates if unsupported and prompted to be uninstalled completely.

Why have this at all if they can't install it and go to the main menu?
image

I would argue that's creating bugs and unused code which was previously working fine.

ExtremeFiretop:
Reply from user:

Post in thread 'MerlinAU v1.4.1 - The Ultimate Firmware Auto-Updater (GNUTON SUPPORT!)' https://www.snbforums.com/threads/merlinau-v1-4-1-the-ultimate-firmware-auto-updater-gnuton-support.91326/post-951249

MerlinAU.sh
Martinski4GitHub: > Could you explain your thinking why this is required when the original design works as it always did by allowing it to install but blocking it from running?
>
>
When we had a single file to install (i.e. only the shell script), it was OK to allow users to download & install the script even for unsupported models, then generate the messages informing them when the current model was unsupported, and ask them to uninstall. It was a valid solution for the simple installation of one shell script file that did not make any changes to the rest of the system.

However, the paradigm has now changed; starting with the **1.4.0** version, we are installing a shell script plus the WebGUI page, the code is modifying the internal system (e.g. the menuTree.js) and adding extra files in the "/www/user" directory when mounting the WebGUI. All these additional changes to the system should be completely avoided when we know **in advance** that the router does not meet the minimum requirements; therefore, the checks must be done before doing the full installation. That’s how the S/W evolves over time to handle and address a more complicated installation; it cannot keep doing what it was doing originally just because "that's the way it used to work before."

Also, it’s considered best practice to avoid making any changes to the system for unsupported H/W, unsupported OS or F/W versions, and unsupported platforms/models. IOW, the goal is to avoid making unnecessary changes to the system in the first place, so we stop the installation process as early as possible, and inform the user of the reason(s) why the S/W cannot be installed.

In the case where users force-install the S/W with some type of hack or workaround, the other existing checks made during execution will prevent the S/W from fully functioning as intended due to known limitations or lack of some minimum requirements. This redundancy is not necessarily a bad thing.

The point is that there's no logical or technical reason to let users fully install the add-on, make changes to the system, and then, once completed, to immediately ask them to uninstall it because it does not support the platform. It’s more efficient, practical, and logical to stop the full installation as early as possible in the first place.

The bug that needed to be fixed was to allow the uninstallation of the shell script itself, and this PR addresses that bug, so users can now type the following command:    

`/jffs/scripts/MerlinAU.sh uninstall`

I hope this explains the reasoning for overhauling the installation process back on the first PRs I submitted for the 1.4.0 version at the beginning of January (about 3 months ago).

ExtremeFiretop: Yes it does. I appreciate the explanation. I think knowing why your changing the flow of things helps a lot. But when something seems technically functional I was always trained we stop touching it unless we need too down the line again for some reason.

The old joke that "there's nothing more permanent than a temporary solution" is both a joke and a fact of life I sometimes take too seriously. I tested the other way, it worked and I get frustrated we were changing the flow of things when it could of been a small adjustment to the flow like I made to resolve the original report.

But understanding why you want to change the flow makes sense and helps frame why the change should be made and the value behind it. Until that point it felt we were almost just changing things without any technical reason or report.

(Which I know bothers you as much as it bothers me.) we do things for reasons and not aimlessly. My change was not aimless and worked exactly as I tested it. But I can see why it's in efficiency and the value in changing the flow when explained clearly as you just did.

ExtremeFiretop:
All good to go! Merging :1st_place_medal:

ExtremeFiretop (approved):

MerlinAU.sh
Martinski4GitHub: > Yes it does. I appreciate the explanation. I think knowing why your changing the flow of things helps a lot. But when something seems technically functional I was always trained we stop touching it unless we need too down the line again for some reason.
> 
> The old joke that "there's nothing more permanent than a temporary solution" is both a joke and a fact of life I sometimes take too seriously. 
>
>
LOL!! I get what you mean. As long as the code works and the functional requirements & technical principles for how it works & behaves remain the same, there's very little reason to change it. However, when new requirements and specifications come in and the "old way" functionally no longer meets those or falls short of best practices, it's time to make changes. If you don't, you are more likely to begin accumulating "technical debt." And any self-respecting S/W developer will tell you: you always want to **reduce** "technical debt" as much as possible, **not** increase it.

>
>
> I tested the current way, it worked and I got frustrated we were changing the flow of things when in my eyes all it needed was a small adjustment to the flow like I made to resolve the original report.
> 
> But understanding why you want to change the flow makes sense and helps frame why the change should be made and the value behind it. Until that point it felt we were almost just changing things without any technical reason or report.
> 
> (Which I know bothers you as much as it bothers me.) we do things for reasons and not aimlessly. My change was not aimless and worked exactly as I tested it with no unintentional behavior which is what I usually associate to a "bug" report.
> 
> But I can see why it's inefficient and the value in changing the flow when explained clearly as you just did.
>
>
Back when I overhauled the installation process to change the workflow for the reasons already explained, I assumed that you either understood the reasons or you were going to ask questions about it. But the questions never came up, so we both just moved forward onto other things. Looking back, it was my fault that I didn't explicitly explain the changes I made, but at the time, there were so many accumulated changes (see PR #384), that it was overwhelming trying to explain every single change that was made.

Anyway, this is another "lesson learned" for both of us, and now we move on having a better understanding of things.

Have a good night, bud!! And have a great weekend!!

ExtremeFiretop: > > The old joke that "there's nothing more permanent than a temporary solution" is both a joke and a fact of life I sometimes take too seriously.
> 
> LOL!! I get what you mean. As long as the code works and the functional requirements & technical principles for how it works & behaves remain the same, there's very little reason to change it. However, when new requirements and specifications come in and the "old way" functionally no longer meets those or falls short of best practices, it's time to make changes. If you don't, you are more likely to begin accumulating "technical debt." And any self-respecting S/W developer will tell you: you always want to **reduce** "technical debt" as much as possible, **not** increase it.
> 

I totally hear that and understand that. This project is allowed to adapt and move. It's not like where I work and where I'm trained. Remember I work in a place with technical debt is just apart of life. The federal is still running cobol in some places. Often the joke really does become a fact of life for us. 


> Back when I overhauled the installation process to change the workflow for the reasons already explained, I assumed that you either understood the reasons or you were going to ask questions about it. But the questions never came up, so we both just moved forward onto other things. Looking back, it was my fault that I didn't explicitly explain the changes I made, but at the time, there were so many accumulated changes (see PR #384), that it was overwhelming trying to explain every single change that was made.
> 

Trust me I didn't expect you to explain every change, but ill admit I just kinda glanced over some of the flow changes and just allowed it assuming there was a reason for it. (Knowing you don't make aimless changes) While I didn't dive deep I probably just assumed it was done for the benefit of the script and the WebUI we were adding.

Obviously there was many changes and this just may have been one of the changes I should of looked a little closer in initially but just brushed over in the originally PR.

> Anyway, this is another "lesson learned" for both of us, and now we move on having a better understanding of things.
> 
> Have a good night, bud!! And have a great weekend!!

You too!  Sorry again.

Martinski4GitHub: > > > The old joke that "there's nothing more permanent than a temporary solution" is both a joke and a fact of life I sometimes take too seriously.
> > 
> > 
> > LOL!! I get what you mean. As long as the code works and the functional requirements & technical principles for how it works & behaves remain the same, there's very little reason to change it. However, when new requirements and specifications come in and the "old way" functionally no longer meets those or falls short of best practices, it's time to make changes. If you don't, you are more likely to begin accumulating "technical debt." And any self-respecting S/W developer will tell you: you always want to **reduce** "technical debt" as much as possible, **not** increase it.
> 
> I totally hear that and understand that. This project is allowed to adapt and change. It's not like where I work and where I'm trained stuff is to stay static. It's a bit of conflicting training. Remember I work in a place with technical debt is just apart of life. The federal is still running cobol in some places. 
>
>
COBOL???  Oh, boy!!!
I haven't seen COBOL code since the mid 1980s - in a book!!!

I've heard also that some departments in the US Federal Govt. are still running some COBOL programs on Windows XP machines. I don't know if that's just some urban myth, or if it's actually true; but I would not be surprised. When I worked in a project we did for the Federal Govt. I was told that they still had some Windows 7 machines running some databases. This was about 6 years ago!!

ExtremeFiretop: Oh I can almost guarantee it's not a myth and is a fact of life in your federal as well. 

Usually you'll find some old air gapped machine running XP or 7 and some old database supporting like all our hydro electric infrastructure or for example  countries financial banking systems. Kinda nuts.

I'm luckily not the one having to deal with those systems directly but I'm told they exist and we have to be aware of them when we make changes on the engineering team.

Usually for those old systems they will bring in someone on contract similar to you with more years in coding than I have alive to hot patch them lol. I'm sure they walk away with lots and lots of money in their pockets at the end of the day.

Screenshot_2025-04-13-06-42-47-32_680d03679600f7af0b4c700c6b270fe7~3

Martinski4GitHub: > Oh I can almost guarantee it's not a myth and is a fact of life in your federal as well.
> 
> Usually you'll find some old air gapped machine running XP or 7 and some old database supporting like all our hydro electric infrastructure or for example countries financial banking systems. Kinda nuts.
> 
> I'm luckily not the one having to deal with those systems directly but I'm told they exist and we have to be aware of them when we make changes on the engineering team that impact the entire department
> 
> Usually for those old systems they will bring in someone on contract similar to you with more years in coding than I have alive to hot patch them lol. I'm sure they walk away with lots and lots of money in their pockets at the end of the day.
> 
>
LOL!!!! Yeah, I hear you!! Sometimes, it's more expedient and convenient to keep an old, but still working sufficiently well, system running than upgrade it to an "unknown" solution which may have its own set of issues. Essentially, the "old" equipment is kept running until "the wheels come off, the rusty nail and screws can't hold it together, and all the people that can maintain it are dead."

PR #480, opened 2025-06-02
Modified RAM Requirements for F/W Flash Process

Martinski4GitHub:
Made changes in the memory requirements for routers with **512MB** RAM to be less restrictive and allow the process to continue with the F/W flash.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,7 +4,7 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2025-May-31
+# Last Modified: 2025-Jun-02
 ###################################################################
 set -u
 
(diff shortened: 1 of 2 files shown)


Martinski4GitHub:
@ExtremeFiretop,

Please take a look and review the PR changes when you have time. The changes mainly affect routers with 512 MB RAM, but I did make slight modifications for routers with 1GB RAM or more.

Essentially, for routers with 512MB RAM, the implementation has 2 fronts:

a) Lower the initial required RAM overhead from 50% to 35% **before** the ZIP file is downloaded, and then lower it again to 25% **after** the F/W image is extracted.

I tested this with my own RT-A86U, and it worked fine, although I did not reach the low memory condition. I also tested with my cousin's router, but I didn't do the full F/W flash as I didn't want to interfere with their home network (one of his teenage kids was playing some game very late at night - summer vacation and out of school - so I was careful not to overstay my welcome).

b) When trying to acquire more available RAM, the code attempts to disable some ASUS/TrendMicro services: 
    AsusNat Tunnel, AiProtection, Traffic Analyzer/Monitor, and Bandwidth Monitor

   This part I did not test with my cousin's router. On my RT-AC86U, I tested with AiProtection services enabled, and it worked to disable the service, kill the processes, flash the F/W image, and reboot.

I'll notify the user @DrEthan77 so that they can test this latest implementation on their own RT-AX86S router.

ExtremeFiretop (approved):

ExtremeFiretop (approved):

Martinski4GitHub:
> I tested it on my Gnuton router since it's the only one with 512Mb of RAM I have. This was the results to see the new RAM requirements: image
> 
> Then I put the memory reporting in debug and forced a specific value to see the process complete which looked like this: image
> 
> Approved!
>
>
Great tests and validation!!

Martinski4GitHub:
> I like the fact that this is a hybrid of my suggestion and your suggestion, to essentially lower the required overhead once the extraction and download is completed.
> 
> But also to shutdown the TrendMicro stuff for memory
>
>
Yes, the implementation attempts to solve the low-memory condition on the 2 fronts. I think the key is to lower the overhead **after** we've already downloaded the ZIP file and extracted its contents, because at that point we've already used quite a bit of the overhead, especially for routers with only 512 MB RAM.

PR #482, opened 2025-06-03
Update README.md: Fixed date string.

Martinski4GitHub:
Fixed date string.

I think it was meant to be June-**03**, unless you have something else coming in for June-**04** ??


ExtremeFiretop:
> Fixed date string.
> 
> I think it was meant to be June-**03**, unless you have something else coming in for June-**04** ??

Too close to Midnight and too late for me to think properly ;)

PR #476, opened 2025-05-19
Fix Web Access Restrictions

ExtremeFiretop:
Fix Web Access Restrictions by allowing subnets shorter than /24

PR #489, opened 2025-06-17
Fix High Risk Matches in Emails

ExtremeFiretop:
Fix Matches in Emails for High Risk Phrases.

Currently the logic only works to highlight the high risk phrases in the email assuming they are all on the same line.
We recently fixed this by flattening the content in PR: https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/476

But only for the verification process, not for the email contents.
This is the issue found below with "Must be manually"
image

This is the fixed HTML version:
image

And the fixed Plain text version:
image

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,1 @@
+blank_issues_enabled: false   # removes the “blank issue” option


ExtremeFiretop:
> Approved!!

Thank you good Sir!

PR #492, opened 2025-06-28
Patch Gnuton Version Detection

ExtremeFiretop:
Patch Gnuton Offline Version Detection


Martinski4GitHub:
> Patch Gnuton Offline Version Detection
> 
>     * **Issues found below in detail:**
> 
>     * Gnuton beta file names are longer and use a different format than our regex currently handles:
>       image
> 
>     * This means when doing an offline flash, the version is not detected as a "BETA" as found below:
>       image
> 
>     * However on Merlin builds the detection does work for alpha and beta builds as found below:
>       image
> 
>     * The end result is the script cannot find the beta file as found below:
>       image
> 
> 
> 2 Fixes introduced:
> 
>     1. Allow the Offline Detection function _GetOfflineFirmwareVersion_ to correctly detect and keep _alpha pr _beta suffix tags
> 
>     2. Adjust the _GetLatestFWUpdateVersionFromGitHub_ function to work with "pre-release" tagged releases instead of only the "latest" release.
> 
> 
> **The end result is found below:** -Stable Release Working: image
> 
> -Beta Releases Now working as well: image
>
>
Very good catch!! It looks like Gnuton keeps throwing a few "curveballs" with the naming of the image files, especially WRT Alpha and Beta releases :>). But I think you nailed it down this time.

Martinski4GitHub (approved):
Approved!! Good job, Bud!!

ExtremeFiretop:
> > Patch Gnuton Offline Version Detection
> > ```
> > * **Issues found below in detail:**
> > 
> > * Gnuton beta file names are longer and use a different format than our regex currently handles:
> >   image
> > 
> > * This means when doing an offline flash, the version is not detected as a "BETA" as found below:
> >   image
> > 
> > * However on Merlin builds the detection does work for alpha and beta builds as found below:
> >   image
> > 
> > * The end result is the script cannot find the beta file as found below:
> >   image
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 2 Fixes introduced:
> > ```
> > 1. Allow the Offline Detection function _GetOfflineFirmwareVersion_ to correctly detect and keep _alpha pr _beta suffix tags
> > 
> > 2. Adjust the _GetLatestFWUpdateVersionFromGitHub_ function to work with "pre-release" tagged releases instead of only the "latest" release.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **The end result is found below:** -Stable Release Working: image
> > -Beta Releases Now working as well: image
> 
> Very good catch!! It looks like Gnuton keeps throwing a few "curveballs" with the naming of the image files, especially WRT Alpha and Beta releases :>). But I think you nailed it down this time.

While I agree that I might of got the regex closer to right, I now wonder why it's required at all.
I had very "productive" sleep last night and feel very rested and woke up immediately going... Why is the offline feature trying to talk to GitHub? Doesn't that go against the whole "offline" intentions?

Now I'm wondering if instead of patching the function if I should be adjusting the actual offline flow... I'll review more today to see why the logic takes the path it takes for an offline flash.

PR #494, opened 2025-07-01
Make Offline Mode True Offline

ExtremeFiretop:
Implemented the new changes for _GetRequiredRAM_KB_  into the offline mode for Gntuon and Merlin firmware

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4946,29 +4946,22 @@ _GetLatestFWUpdateVersionFromWebsite_()
 }
 
 ##------------------------------------------##
-## Modified by ExtremeFiretop [2025-Jun-28] ##
+## Modified by ExtremeFiretop [2024-May-05] ##
 ##------------------------------------------##
 _GetLatestFWUpdateVersionFromGitHub_()
 {
-    local routerVersion  search_type
+    local routerVersion
     local gitURL="$1"  # GitHub URL for the latest release #
     local firmware_type="$2"  # "tuf", "rog" or "pure" #
-    local grep_pattern downloadURLs theURL urlVersion
 
-    search_type="$firmware_type"  # Default to the input firmware_type #
+    local search_type="$firmware_type"  # Default to the input firmware_type #
 
-    # If firmware_type is "pure", set search type to include "squashfs" as well #
+    # If firmware_type is "pure", set search_type to include "squashfs" as well
     if [ "$firmware_type" = "pure" ]
     then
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
This is what I should of initially done, but you know me, I go with the quickest and easiest fix that addresses the problem head on.
But like I mentioned, after a good nights sleep I reflected and realized if I was a user of the offline mode, and I saw an error that it couldn't find the URL after I specifically uploaded the ZIP, i would be confused and it wouldn't make any sense.

It's why I dived into the deeper logic on why offline mode wasn't actually offline in the first place. The truth is there is no real reason it couldn't be truly offline in offline mode.

Martinski4GitHub:
> @Martinski4GitHub
> 
> Things done in this PR:
> 
>     1. Restored the _GetLatestFWUpdateVersionFromGitHub_ function to only work with production releases (no longer finds pre-releases, same as production now before my latest PR: [Patch Gnuton Version Detection #492](https://github.com/ExtremeFiretop/MerlinAutoUpdate-Router/pull/492) )
> 
>     2. Implemented your changed _GetRequiredRAM_KB_ function in the offline mode
> 
>     3. Finally; Modified the _RunOfflineUpdateNow_ to remove the steps currently in production which still require a "online connection" to an external source
> 
> 
> Now the offline mode is truly offline, and the require memory calculations are still working as expected as found below:
> 
>     * **GNUTON OFFLINE:**
>       image
> 
>     * **RMERLIN OFFLINE:**
>       image
>
>
This is looking good. And you're right, the "Offline F/W Update" mode should really be completely "offline."

Martinski4GitHub:
> This is what I should of initially done, but you know me, I go with the quickest and easiest fix that addresses the problem head on. But like I mentioned, after a good nights sleep I reflected and realized if I was a user of the offline mode, and I saw an error that it couldn't find the URL after I specifically uploaded the ZIP, i would be confused and it wouldn't make any sense.
> 
> It's why I dived into the deeper logic on why offline mode wasn't actually offline in the first place. The truth is there is no real reason it couldn't be truly offline in offline mode.
>
>
This is a much "cleaner" implementation for the Offline F/W Update. As you said, it really should not need to go out and remotely retrieve information because, presumably, the user doing the "Offline Update" should know what he's doing and would provide the required file(s) locally to perform the update.

It was a good call to review the workflow and come up with a better & cleaner solution. Good job, bud!! You need to get more good night's sleep :>)

PR #497, opened 2025-07-18
Cleaning Up Temporary Unused Code

ExtremeFiretop:
1. Cleaning Up Temporary Unused Code
2. Increasing the Minimum Supported Firmware to 3004.388.8.4 for 3004_388 based routers


Martinski4GitHub:
> 1. Cleaning Up Temporary Unused Code
> 
>     2. Increasing the Minimum Supported Firmware to 3004.388.8.4 for 3004_388 based routers
>
>
Good thing that you're keeping track of these items because, honestly, those were not on my radar at all, LOL!!!

P.S. 
Just FYI, my wife and I are about to start watching a movie, so I'll be offline for a couple of hours.

ExtremeFiretop:
> 
> Good thing that you're keeping track of these items because, honestly, those were not on my radar at all, LOL!!!
> 

There is no harm in keeping the code of course. 

But I would argue it's purpose has dried up, and it's time for it to be removed as it no longer plays any valid role. Both these chunks or code were originally designed to help ease over the move to 1.4.0 with the WebUI, and now the code sits unused and can be cleaned up.

If somehow someone is still on 1.3.9 without the WebUI they will have issues updating to 1.5.1 but at that point I would argue it's on them.

ExtremeFiretop:
> I already understood the reasons based on previous instances, and I'm not opposed to the latest cleanup if that's what you thought from reading my previous message.

LOL Honestly no I didn't think that I was mostly talking out loud, patting myself on the back and appreciating my genius for deleting a few lines ;) HAHAHA

> Ah, yes, of course - but you're preaching to the choir!!😉 

True, but you guys have the best acoustics. ;) 
As long as im not preaching to a brick wall. I’ll take an amen over an echo any day. LOL

PR #499, opened 2025-07-29
Correct sed(1) RE to extract version from the Gnuton FW filename.

cdavis5e:
So far, I've seen `ubi`, `nand_squashfs`, and `puresqubi`. While I'm at it, tighten up the REs.

Fixes updating models such as the ZenWiFi XT8 (RT-AX95Q) that use `puresqubi` or `nand_squashfs` Gnuton images.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4996,7 +4996,7 @@ _GetLatestFWUpdateVersionFromGitHub_()
             # Extract the version portion from the URL #
             urlVersion="$(echo "$theURL" \
                 | grep -oE "${PRODUCT_ID}_[^ ]*\.(w|pkgtb)" \
-                | sed "s/${PRODUCT_ID}_//;s/.w$//;s/.pkgtb$//;s/.ubi$//;s/_/./g;s/-gnuton[0-9][0-9]*\$//" | head -n1)"
+                | sed "s/${PRODUCT_ID}_//;s/\\.w$//;s/\\.pkgtb$//;s/_\\(ubi\\|puresqubi\\|nand_squashfs\\)$//;s/_/./g;s/-gnuton[0-9][0-9]*$//" | head -n1)"
 
             if [ "$urlVersion" = "$routerVersion" ]
             then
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
> So far, I've seen `ubi`, `nand_squashfs`, and `puresqubi`. While I'm at it, tighten up the REs.
> 
> Fixes updating models such as the ZenWiFi XT8 (RT-AX95Q) that use `puresqubi` or `nand_squashfs` Gnuton images.
>
Great catch WRT the different file extensions!!

Martinski4GitHub:
> > > In the function of: _GetLatestFWUpdateVersionFromGitHub_ My PR: would make the variable: routerVersion=3004.388.9.2-gnuton1
> > > But if I merge this PR in as is; then the variable: urlVersion would return without the -gnuton1 suffix, and the final match would fail: if [ "$urlVersion" = "$routerVersion" ]
> > > @cdavis5e can you also modify this so it wouldn't conflict with the upcoming planned PR to return the long version (-gnuton1 suffix included)
> > 
> > 
> > Yeah, that would be the best way to integrate the fixes in this PR.
> 
> It's a small tweak to make, but otherwise I agree. It's a good catch and surprised I missed it before.
> 
> It's clear to me because I'm only testing with a ubi router from Gnuton. Perfect example of why "it works for me" doesn't always cut it 😜
>
>
Exactly, LOL!!!  A similar situation happened with scMerlin's "WAN Uptime" feature, where we missed the "Dual WAN with Failover" setup because we don't have it to test and validate with. That's why Beta testing with external users is so critical.

ExtremeFiretop (changes requested):
Changes Required to not conflict with PR: 498

Martinski4GitHub:
> @Martinski4GitHub
> 
> Seems to me that @cdavis5e opened a PR and then disappeared? lol I had responded to him within no more than half an hour when this was opened!
> 
> It's now been 4 days with no movement so I made my own PR: #501
> 
> And was close this one.
>
>
Yeah, that's fair. There was plenty of time to reply, even if it had been just a simple: "I don't have time at the moment."
In any case, the poster will get their due credit.

PR #511, opened 2025-08-14
Don't purge the configuration file early

ExtremeFiretop:
Don't purge the configuration file early

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -9568,7 +9568,6 @@ _CheckForMinimumRequirements_()
 
    "$requirementsCheckOK" && return 0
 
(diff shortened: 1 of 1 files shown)


MerlinAU.sh
```diff
@@ -9568,7 +9568,6 @@ _CheckForMinimumRequirements_()
 
    "$requirementsCheckOK" && return 0
 
-   rm -f "$CONFIG_FILE"
```
Martinski4GitHub, line 9571: Yep, that should work fine. Good call!!

Martinski4GitHub (approved):
Approved!!

PR #525, opened 2025-11-06
Fix for "parameter not set" Error Message

Martinski4GitHub:
Fixed a bug generating a "currentDIVER_version: parameter not set" error message.


ExtremeFiretop:
> > > So I reviewed the script and found the root cause around the code that handles stopping and restarting Diversion. This bug is now fixed with this PR.
> > > Now, I'm not quite sure if making another production release is warranted for this bug fix since this scenario is probably rare, although it does affect restarting Diversion when the error happens.
> > > What do you think?
> > 
> > 
> > I guess the question is how rare is this scenario? Can you describe it in further detail?
> > I'm on my phone currently so my ability to analyze this is limited but I can see the code got moved up in the script.
> > I also know that the last time I tested MerlinAU 1.5.5 it did infact shutdown diversion for me as expected. So I would need to better understand the scenario that causes this to assess how quickly we should pump out a release.
> 
> Based on my review of the code, the scenario requires **Diversion** to be installed **and** a **failure** to log into the router WebUI. In my friend's case, it failed because of a bad password, but I suppose it could also fail for some other reason, but it may be an unusual failure.
> 

Okay I better understand the situation now, I would have to agree that's pretty rare, but something that deserves to be flagged and fixed. I've approved the code changes in the other PR so you can merge in as you see fit.

> Yeah, I understand. I tend to wait a little bit to get verification from a 3rd-party, and also based on the severity of the problem. For example, if it's a consistent crash that happens, I definitely would make a release right away.
> 

If you feel like waiting until I return, I should be back this coming Monday, but I've approved the changes preemptively so you have nothing stopping you now.

> > However as I joked in the forums, don't feel like you need my approval, you have the keys to the kingdom while I'm away and I trust you and your judgement 100%
> 
> Well, I only have access to half the kingdom, LOL!!

Yes, seems like that's something else we need to investigate, I appreciate you flagging, half a key is not going to unlock a door, if I'm giving you keys while I'm gone I need to give you the full key 😂

PR #531, opened 2025-11-11
Email Notification when USB Drive Unmount Fails

Martinski4GitHub:
Added an email notification when unmounting the USB-attached drive fails after trying for 4 minutes.


ExtremeFiretop (approved):

ExtremeFiretop:
I like this idea. Give users the FYI that their USB can't be unmounted for some reason

PR #546, opened 2026-02-07
Save Most Recent Log File

Martinski4GitHub:
Added code to save the most recent F/W update log file. This way at least the most recent remains available to review when deleting all log files older than 30 days.


ExtremeFiretop (approved):

ExtremeFiretop:
Smart call, realistically I expect that most of the time, it would keep at least 30 days of **daily** logs.
However, it's possible the script is set to run manually, or monthly by cron job, in which case the "delete anything older than 30 days" would delete everything.

PR #549, opened 2026-02-22
Support for New AMTM Functionality

Martinski4GitHub:
- Initial code to support automatic script updates via AMTM.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,16 +4,16 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2026-Feb-07
+# Last Modified: 2026-Feb-22
 ###################################################################
 set -u
 
 ## Set version for each Production Release ##
-readonly SCRIPT_VERSION=1.5.9
-readonly SCRIPT_VERSTAG="26020700"
+readonly SCRIPT_VERSION=1.5.10
+readonly SCRIPT_VERSTAG="26022202"
 readonly SCRIPT_NAME="MerlinAU"
 ## Set to "master" for Production Releases ##
-SCRIPT_BRANCH="master"
+SCRIPT_BRANCH="dev"
 
 ##----------------------------------------##
 ## Modified by Martinski W. [2024-Jul-03] ##
@@ -51,6 +51,9 @@ scriptUpdateNotify=0
 # For router model check #
 routerModelCheckFailed=false
 offlineUpdateTrigger=false
+
+# To support automatic script updates from AMTM #
+doScriptUpdateFromAMTM=true
 
 ##----------------------------------------##
 ## Modified by Martinski W. [2025-Feb-18] ##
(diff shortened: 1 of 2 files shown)


ExtremeFiretop:
> @ExtremeFiretop,
> 
> This PR contains changes & additions to support automatic script updates via AMTM. I have tested as much as I could, but I would not mind a 2nd pair of eyes, so whenever you get a chance, please take a look and validate. There's no hurry, so take your time.
> 
> I hope you're having a great weekend, bud.

Wow buddy! Thanks so much! 
I appreciate the fact that you took the initiative to help here to get it started. 

I myself was more or less waiting for your word on OSR before I started with MerlinAU, but it seems you already got a big chunk of OSR done as well!

Still, if you want me to take on any of the OSR add-ons let me know.

I'll test these changes today, and report back or submit any changes if I notice anything.

Hope you and the wife also have a great weekend! Enjoy some good movies for me 😉 Any recent movie hits I should consider adding to my playlist?

ExtremeFiretop:
Question for you @Martinski4GitHub 

How do you feel about this request?
https://www.snbforums.com/threads/merlinau-v1-5-9-the-ultimate-firmware-auto-updater.96306/post-985278

ExtremeFiretop:
> > Question for you @Martinski4GitHub
> > How do you feel about this request? https://www.snbforums.com/threads/merlinau-v1-5-9-the-ultimate-firmware-auto-updater.96306/post-985278
> 
> I took a brief look at the website (https://docs.ntfy.sh/), and it seems interesting. However, I don't like that the "topics" (i.e. channels), one subscribes to are completely public, unsecured, and unprotected from any prying eyes, so for that alone I wouldn't use this service for my own personal notifications, and I don't know if they are collecting/mining data from the notifications themselves (I'm not too paranoid, but paranoid enough... 😉).
> 
> OTOH, if the notifications being sent have no personally identifiable information, and no other type of data that the user might consider "private" - none whatsover - in their own eyes, I suppose the service could be used for MerlinAU notifications with complete disregard to security and/or privacy.
> 
> Now, I know the documentation has instructions to self-host your own **ntfy** server and create a fully private notification channel, but most SNB Forum users don't appear to be tech-savvy enough to do this on their own, so they would follow "the path of least resistance."

Totally agree with all the privacy points, I myself also would not consider such a service, but I think we would lose that argument due to the fact that the notifications themselves can be made with no personally identifiable information (as you mentioned) or host that a private instance.

> 
> The setup and actual call to the service appear fairly simple, so I don't think that part of the implementation is a major undertaking. I suspect most of the work would involve modifying existing code to "reformat & redirect" all the current email notifications to the new notification service.
> 
> WRT options, I imagine users will want to configure one of four scenarios:
> 
> 1. Disable ALL notifications
> 2. Send Notifications ONLY via Email
> 3. Send Notifications ONLY via Ntfy
> 4. Send Notifications via BOTH Email and Ntfy
> 
> Now, I expect that adding this new notification service may open the door for requests to add support for **Pushover** and maybe even **Discord** as well... 🤷🏻‍♂️

This is the real show stopper for me. I don't want to start to integrate into every service imaginable. 
We do this, and we open the door to everything else. Which is my real concern and the main reason I'm not overly a fan of the idea.

> 
> To make a long story short: I'm not completely opposed to adding this **ntfy** service to MerlinAU, but I do have my own personal reservations, so it's not something I would recommend to family or close friends with full confidence.
> 
> Just my 2 cents.

We are in agreeance then for now. No need to investigate more for the time being. :)

PR #568, opened 2026-05-07
Fix for Email Notifications for New F/W Updates

Martinski4GitHub:
Fixed a bug where sometimes an email notification was not sent when a new F/W update version was detected.
Thanks to @ExtremeFiretop  for finding the bug!!

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -4,16 +4,16 @@
 #
 # Original Creation Date: 2023-Oct-01 by @ExtremeFiretop.
 # Official Co-Author: @Martinski W. - Date: 2023-Nov-01
-# Last Modified: 2026-Apr-08
+# Last Modified: 2026-May-06
 ###################################################################
 set -u
 
 ## Set version for each Production Release ##
-readonly SCRIPT_VERSION=1.6.1
-readonly SCRIPT_VERSTAG="26040823"
+readonly SCRIPT_VERSION=1.6.2
+readonly SCRIPT_VERSTAG="26050619"
 readonly SCRIPT_NAME="MerlinAU"
 ## Set to "master" for Production Releases ##
-SCRIPT_BRANCH="master"
+SCRIPT_BRANCH="dev"
 
 ##----------------------------------------##
 ## Modified by Martinski W. [2024-Jul-03] ##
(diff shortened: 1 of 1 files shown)


Martinski4GitHub:
@ExtremeFiretop,

Well, I became extremely curious about what might be going wrong WRT the "**FW_New_Update_Notification_Date**" parameter setting causing issues with email notifications, so I decided to take a quick look after my short "power nap" and while waiting for dinner to be ready.

I reviewed the main function where all the related "F/W update" settings are handled, and, rather quickly, I found the problem in the "`_CheckNewUpdateFirmwareNotification_`" function.

I have tested & verified the fix, so please take a look whenever you get a chance and double-check it. There's no need to reset the "FW_New_Update_Notification_Date" parameter setting too early (i.e. during post-reboot right after the F/W Update has been completed).

Excellent catch, bud!!

MerlinAU.sh
```diff
@@ -5420,6 +5420,11 @@ _GetLatestFWUpdateVersionFromWebsite_()
     then echo "**ERROR** **NO_URL**" ; return 1
     fi
 
+    ## Make sure "386.xx" versions have '3004.' prefix ##
+    if echo "$versionStr" | grep -qE '^386[.][0-9]+'
+    then versionStr="3004.$versionStr"
+    fi
```
Martinski4GitHub, line 5426: This change was needed to run the code and test on older AC-class routers with **3004.386.XX** F/W versions. Yes, they are EOL, but for testing purposes, it's very useful to have the code working.

MerlinAU.sh
```diff
@@ -6669,7 +6674,7 @@ _Calculate_NextRunTime_()
         if [ -z "$fwNewUpdateNotificationDate" ] || \
            [ "$fwNewUpdateNotificationDate" = "TBD" ]
         then
-            fwNewUpdateNotificationDate="$(date +%Y-%m-%d_%H:%M:%S)"
+            fwNewUpdateNotificationDate="$(date +"$FW_UpdateNotificationDateFormat")"
```
Martinski4GitHub, line 6677: To keep the date format consistent.

MerlinAU.sh
```diff
@@ -8334,6 +8339,8 @@ _CheckNewUpdateFirmwareNotification_()
    then
        fwNewUpdateNotificationVers="$releaseVersionStr"
        Update_Custom_Settings FW_New_Update_Notification_Vers "$fwNewUpdateNotificationVers"
+       # Reset for newer F/W version updates #
+       Update_Custom_Settings FW_New_Update_Notification_Date TBD
```
Martinski4GitHub, line 8343: This is the fix to make sure the setting is reset every time a new F/W version update is detected.

PR #572, opened 2026-05-21
Bug-Fix to Update Process

ExtremeFiretop:
Reported in the forums here: https://www.snbforums.com/threads/merlinau-v1-6-2-the-ultimate-firmware-auto-updater.96306/post-993395


Martinski4GitHub:
@ExtremeFiretop,

You had the right idea for the fix. After a closer review, I decided to make code improvements and fine-tune some areas to make the script much more robust when checking for and running in a non-interactive shell.

See my review comments in PR #573 for some explanations of my modifications/additions.

That was a great catch by user 'scootertramp' from the SNB Forums.

PR #581, opened 2026-06-22
Fix log cleanup function to create directory and restore logs

ExtremeFiretop:
Ensure log directory exists before cleanup and enhance log deletion with restoration attempt.

### DIFF
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -9,11 +9,11 @@
 set -u
 
 ## Set version for each Production Release ##
-readonly SCRIPT_VERSION=1.6.4
-readonly SCRIPT_VERSTAG="26061118"
+readonly SCRIPT_VERSION=1.6.5
+readonly SCRIPT_VERSTAG="26062210"
 readonly SCRIPT_NAME="MerlinAU"
 ## Set to "master" for Production Releases ##
-SCRIPT_BRANCH="master"
+SCRIPT_BRANCH="dev"
 
 ##----------------------------------------##
 ## Modified by Martinski W. [2024-Jul-03] ##
--- a/MerlinAU.sh
+++ b/MerlinAU.sh
@@ -2250,7 +2250,7 @@ readonly POST_UPDATE_EMAIL_SCRIPT_HOOK="[ -x $ScriptFilePath ] && $POST_UPDATE_E
 ##------------------------------------------##
 _CleanUpOldLogFiles_()
 {
-    [ ! -d "$FW_LOG_DIR" ] && return 1
+    [ ! -d "$FW_LOG_DIR" ] && mkdir -p -m 755 "$FW_LOG_DIR"
     local numLogFiles  topLogFile  savedTopLogFile=""
 
     numLogFiles="$(ls -1lt "$FW_LOG_DIR"/*.log 2>/dev/null | wc -l)"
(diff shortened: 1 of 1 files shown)


ExtremeFiretop:
@Martinski4GitHub 

New installs now seem to create this warning message:
image

The cause seems to be that the log directory is not yet recreated on first install yet.
Simple solution is to create the directory if missing.

Martinski4GitHub:
> @Martinski4GitHub
> 
> New installs now seem to create this warning message: image
> 
>
>
Nice catch!!!

>
>
> The cause seems to be that the log directory is not yet recreated on first install yet. Simple solution is to create the directory if missing.
>
>
See my review comments.

MerlinAU.sh
```diff
@@ -2250,7 +2250,11 @@ readonly POST_UPDATE_EMAIL_SCRIPT_HOOK="[ -x $ScriptFilePath ] && $POST_UPDATE_E
 ##------------------------------------------##
 _CleanUpOldLogFiles_()
 {
-    [ ! -d "$FW_LOG_DIR" ] && return 1
+    # Create the log directory on a fresh installation.
+    if [ ! -d "$FW_LOG_DIR" ]
+    then
+        mkdir -p -m 755 "$FW_LOG_DIR" 2>/dev/null || return 1
```
Martinski4GitHub, line 2256: There's no need or benefit to create the directory at this particular point since the goal of the function is to "**clean up**" old log files, but if there's no log directory to be found, we can simply just do `return 0` because there's actually **nothing to clean up**.

MerlinAU.sh
```diff
@@ -2270,7 +2274,18 @@ _CleanUpOldLogFiles_()
...
+    if ! /usr/bin/find -L "$FW_LOG_DIR" \
+        -name '*.log' \
+        -mtime +30 \
+        -exec rm -f {} \;
+    then
+        # Attempt restoration before returning failure.
+        [ -n "$savedTopLogFile" ] &&
+            [ -e "$savedTopLogFile" ] &&
+            mv -f "$savedTopLogFile" "$topLogFile" 2>/dev/null
+
```
Martinski4GitHub, line 2286: This code is not needed here since it's essentially trying to do the same restoration that the code just below it is doing, so it's duplicated code.

Martinski4GitHub (approved):
I'll merge this PR, but I'll submit another one to clean things up and improve the code.

MerlinAU.sh
ExtremeFiretop: Obviously there's nothing to clean up, but the directory should be created while initialization happens since it's supposed to exist.

Reporting "nothing to clean up" means that it will also report all good later down the line if the it actually had issues deleting the files. I would rather see the error message that it was unable to cleanup as designed in those future instances. 

Changing it to return 0 - would break that functionality and protection.

MerlinAU.sh
ExtremeFiretop: Well yes, but with different purposes /difference paths.

Once is on the failure path, one is on the success path. Without the additional code in the failure path, it would just fail without reaching the restore.

You can re-jig if you want so it's less duplicated and just passes on the return status, but either way the the outcome should be the same. Restoration should happen when calling a failure for whatever reason

Martinski4GitHub: > Well yes, but with different purposes /difference paths.
> 
> Once is on the failure path, one is on the success path. Without the additional code in the failure path, it would just fail without reaching the restore.
> 
>
I know you re-evaluated the situation and changed your mind later on, but I got to tell you, after reading this, I was scratching my head for a while trying to figure out what you were trying to say, LOL!!! 😜😉

In any case, glad we both agreed and came to the same conclusion.

Thanks, bud.

ExtremeFiretop: > > Well yes, but with different purposes /difference paths.
> > Once is on the failure path, one is on the success path. Without the additional code in the failure path, it would just fail without reaching the restore.
> 
> I know you re-evaluated the situation and changed your mind later on, but I got to tell you, after reading this, I was scratching my head for a while trying to figure out what you were trying to say, LOL!!! 😜😉
> 
> In any case, glad we both agreed and came to the same conclusion.
> 
> Thanks, bud.

I'm not even sure what I was trying to say. 🤣
I was tired and for some reason had some other idea of when this function ran and it's purpose.
I read the function name and tested and realized I was asleep at the wheel.

PR #589, opened 2026-07-31
Removal of Cron Jobs with Timeout

Martinski4GitHub:
Modified code that removes cron jobs from 3rd-party add-ons so that the `cru` command call has a timeout to prevent stuck commands.


Martinski4GitHub:
@ExtremeFiretop,

I saw the problem report on the SNB Forums about the `cru` command getting "stuck" while removing a cron job, so I decided to add a simple routine to prevent such a "stuck" command. 

I don't know the reason for the `cru` command getting stuck, and instead of going into a "rabbit hole" or getting lost in the weeds investigating, for now I prefer to avoid the problem altogether, so at least the router is not left in a "hung" state.

Please take a look at the PR changes.

ExtremeFiretop:
> 
> I don't know the reason for the `cru` command getting stuck, and instead of going into a "rabbit hole" or getting lost in the weeds investigating
.

I feel this is targeted at me 😁 I was going to spend some time trying to get some theories tested. But I agree in the meantime just a simple escape is the best solution

Martinski4GitHub:
> > I don't know the reason for the `cru` command getting stuck, and instead of going into a "rabbit hole" or getting lost in the weeds investigating
> > .
> 
> I feel this is targeted at me 😁 I was going to spend some time trying to get some theories tested. But I agree in the meantime just a simple escape is the best solution
>
>
At least for the next production release, the first priority (IMO) is to prevent the "hung" state. For the long run, we could spend some time investigating further what the issue is with the `cru` command, but as I mentioned, that might be a deep "rabbit hole" for which I don't have time at the moment.

I do recall in the past there were some issues with cron jobs going missing during reboots, and I believe that was fixed with an exclusive FLock. Perhaps there is some race condition somewhere that causes a deadlock??

Anyway, I'm being summoned for dinner, so talk to you later, bud.

